Issue #20, opened 2014-08-01
Test failures in environment with Financial Force packages installed


adtennant:
When deploying into a fresh developer edition, all tests pass and everything is hunky dory.

When deploying into an environment which has the followingFinancial Force packages installed (and no other Apex), SObjectSelectorTests fail.

* FinancialForce Expansion Pack (ffxp)
* FinancialForce QuickStart (ffqs)
* FinancialForce ClickLink (ffirule)
* FinancialForce Accounting (c2g)
* FinancialForce Reporting (ffr)
* FinancialForce Accounting Accounts Payable Extension (ffap)
* Basic SCM Reports
* FinancialForce Central (ffc)
* FinancialForce Journal Extension (ffgl)
* FinancialForce Cash Entry Extension (ffcash)
* FinancialForce VAT Reporting (ffvat)
* FinancialForce OneTouch (ffbext)
* SCM (SCMC)
* Declarative Lookup Rollup Summaries Tool (dlrs)
* FinancialForce Enquires 
* FinancialForce Consulting Tools (ffct)

### Test Failures

**Class**
fflib_SObjectSelectorTest
**Method Name**
testDefaultConfig
**Pass/Fail**
Fail
**Error Message**
System.AssertException: Assertion Failed: Expected: AccountNumber,AnnualRevenue,Id,Name, Actual: Name,AnnualRevenue,AccountNumber,Id
**Stack Trace**
Class.fflib_SObjectSelectorTest.testDefaultConfig: line 185, column 1

**Class**
fflib_SObjectSelectorTest
**Method Name**
testGetFieldListString
**Pass/Fail**
Fail
**Error Message**
System.AssertException: Assertion Failed: Expected: myprefix.AccountNumber,myprefix.AnnualRevenue,myprefix.Id,myprefix.Name, Actual: myprefix.AccountNumber,myprefix.Name,myprefix.Id,myprefix.AnnualRevenue
**Stack Trace**
Class.fflib_SObjectSelectorTest.testGetFieldListString: line 46, column 1

up2go-rsoesemann:
Any updates on this? We really want to use the fflib packages but this makes them a risk for every release.

tfuda:
I believe I have fixed this in pull request #33. Because the field collection is backed by a Set, the are returned in an unpredictable order. Since I don't believe there's any particular need to preserve the ordering of the fields in the generated SOQL, I fixed the test assertions to tolerate the fields coming back in any order.

[...]

PR #21, opened 2014-08-02
Update fflib_SObjectSelectorTest.cls

adtennant:
Fixes issue #20, I've not yet figured out why it does, but it does.


afawcett:
@capeterson any update on updating queryfactory to be deterministic?

afawcett:
@capeterson Has this version of queryfactory been updated to be deterministic as per our internal repo?

frup42:
Looks like this was fixed without black magic in #33, closing PR.

PR #25, opened 2014-08-18
Fixed issue 24...

up2go-rsoesemann:
Fixes Issue https://github.com/financialforcedev/fflib-apex-common/issues/24 (fflib_QueryFactory().selectFieldSet() fails with when FieldSet queries
namespaced fields"

Root cause: Field namespaces are internally stored as lowercase map key but not converted when passed in
Solution: add .toLowerCase()


afawcett:
@capeterson This looks ok to me, the platform is case insensitive when it comes to object and field names. I'll merge, but do let me know if you think of something.

PR #44, opened 2015-05-18
CurrencyIsoCode Override update

ImJohnMDaniel:
Added ApexClass and Attachment sObjects to the list of CurrencyIsoCode override string set of STANDARD_WITHOUT_CURRENCYISO


afawcett:
Thanks John!

PR #46, opened 2015-05-26
Fix missing variable referenced in test method

loganm:
The registerDeleted test method was failing because it referred to a variable called 'record' that wasn't present in the scope. I've updated the reference to 'records' which is in scope.


afawcett:
Thanks, really.... need to get CI setup on this repo soon... :+1:

loganm:
No worries @daniel-huckins. Let's just blame it on the fact that Andrew hasn't set CI up on this repo yet ;)

PR #48, opened 2015-06-01
Update to fix deploy, and remove CurrencyIsoCode from RecordType standard object

daveerickson:
Line 224 was referencing 'record' variable instead of 'records' so would not deploy.


daveerickson:
Also added change to not automatically add CurrencyIsoCode to RecordType queries.

afawcett:
Thanks, hope your enjoying using the patterns! :+1:

PR #60, opened 2015-07-02
Added support for reflective lookups

loganm:
Basically, after the insert, if there were reflective relationships specified, I resolve those lookups and then update the records where those reflective relationships were specified.

I ensured that only the records that needed the reflective relationships to be resolved were updated after the insert to minimise the dml rows used.

This feels like a better approach than trying to order batches of inserts of the same SObjectType. The trade off is DML rows vs DML statements, and we have a lot more DML rows to play with.

### DIFF
--- a/fflib/src/classes/fflib_SObjectUnitOfWork.cls
+++ b/fflib/src/classes/fflib_SObjectUnitOfWork.cls
@@ -2,22 +2,22 @@
  * Copyright (c), FinancialForce.com, inc
  * All rights reserved.
  *
- * Redistribution and use in source and binary forms, with or without modification, 
+ * Redistribution and use in source and binary forms, with or without modification,
  *   are permitted provided that the following conditions are met:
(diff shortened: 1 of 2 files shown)


loganm:
Also, sorry about all the unrelated changes, my sublimetext does a lot of cleanup on save.

loganm:
There is another way to achieve this, and that is to go into a loop inserting batches of records where the record id is null and all the related records have Ids. This may be a better approach if some trigger action occurs on insert that depends on the related field. The downside is it will cause an undetermined number of DML statements.

loganm:
There is a bug caused by multiple instances of the same record in the update. I've closed the pull request while I work it out.

afawcett:
Thanks for this, appreciate it.

Don't worry about white space, if you add ?w=1 on the end of the GitHub URL in diff mode it removes it. :+1:

PR #61, opened 2015-07-09
Fixes #57. Exposes registerWork and registerEmail in interface

### LINKED ISSUE #57: Adding work to UOW breaks factory pattern
Using the UnitOfWorkFactory to obtain a fflib_ISObjectUnitOfWork, unable to registerWork as the registerWork/registerEmail are implemented on fflib_UnitOfWork but not exposed via fflib_ISObjectUnitOfWork.  The ability to process "post-DML but pre-commit" operations is incredible useful, would be able to keep consistent use of other patterns if these methods were added to the interface fflib_ISObjectUnitOfWork.

cirotix:
Expose registerWork and registerEmail so that you can still use the Factory Pattern.

### DIFF
--- a/fflib/src/classes/fflib_IDoWork.cls
+++ b/fflib/src/classes/fflib_IDoWork.cls
@@ -30,4 +30,4 @@
 public interface fflib_IDoWork 
 {
     void doWork();
-	}
+}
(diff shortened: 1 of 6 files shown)


afawcett:
Thanks for this. :+1: 

Can i ask why you move IDoWork?

Reason i ask is it breaks backwards compatibility for anyone upgrading who has previously implement it. Yes they could change their code, but we try to avoid this. Hence my question.

afawcett:
Ah sorry, just added to the conversation over on issue 57. :+1:

jondavis9898:
I think this can be closed as it's covered in PR #67

Issue #56, opened 2015-06-28
Improve ability to customize functionality


jondavis9898:
Coming from a non-SFDC background, the fflib framework and enterprise patterns that it promotes are a breath of fresh air.  However, the ability to customize the implementation is difficult because a large number of classes/methods/etc. are not marked virtual.  The solution for this is:

1) Modify the fflib classes directly - Downside is this impacts upgrade pathing
2) Create custom classes that encapsulate the fflib classes and use the custom classes as this - Downside is that fflib classes are still directly available for use and requires "more" code

Would like to see the library use more liberal use of virtual, protected, etc.

afawcett:
Thanks for this @jondavis9898, do you have some use cases or examples that prompted this thought?

jondavis9898:
Why yes I do :)  I'm working on some requirements that need to expand the base functionality for specific scenarios.  I could fork and customize but I'm trying to keep underlying details "abstracted" to ease support/maintenance in the future and since the changes I need to make are relatively minor.  By being able to override methods (e.g. commitWork), I can tweak things just slightly.  Currently, the only viable approach is to directly modify fflib classes which makes upgrade pathing as fflib adds new features difficult.

Here's some use cases I've worked on just in the last couple of days:

1) Automatic "Forced (developer mode) rollups" - In fflib_SObjectUnitOfWork, implement automatic rollup triggering (using DLRS Developer mode) after DML but before IDoWork is processed in commitWork.  To do this, need to add a new member variable to class and override commitWork, howevre unable to do either due to protection levels.  Additionally, check validation errors from domain classes and throw exception (see #2 for more details).  We have record counts of more than 200 so doing rollups this way (instead of using triggerHandler keeps DML statements, queries, etc. to a minimum).
2) Keep Track of "Validation Errors" based on Record Id - In fflib_SObjectDomain, implement a "ValidationErrors" static variable similar to Errors.  Given our record counts, validation must occur after all DML.  This class is virtual but all inner pieces are private and/or not virtual (e.g. can't re-use re-use the FieldError, ObjectError, etc.).
3) "Before Validations" - In fflib_SObjectDomain, need to do some "before validations" and currently no method exposed for this (before only calls onApplyDefaults/onBeforeInsert/Update, etc.).  All that's needed is to override handleBeforeInsert & handleBeforeUpdate but protection levels prohibit.  

[...]

PR #62, opened 2015-07-18
Issue #56 - Add event notifications to commitWork processing

jondavis9898:
Added event notifications during commitWork at various stages of the process. 

Decided to have discrete "Starting/Finished" for each "operation" to support forward compatibility if other "operations" are added to commitWork.  For example, if something is added in-between DML & DoWork in the future and code was written in onDoWorkStarting assuming it was fired immediately after DML was complete it would lead to unexpected behavior.

Kicked around the idea of having a "Successful" and "Failed" notification instead of just a "Finished(flag)" notification but wanted to keep things as simple as possible while still providing the desired outcome.

Appreciate any feedback to further improve.

### DIFF
--- a/fflib/src/classes/fflib_Application.cls
+++ b/fflib/src/classes/fflib_Application.cls
@@ -62,6 +62,22 @@ public class fflib_Application
 			return new fflib_SObjectUnitOfWork(m_objectTypes);
 		}
 
+		/**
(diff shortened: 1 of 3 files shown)


afawcett:
Very nice work indeed! Thanks! :+1:

Issue #58, opened 2015-07-02
Performance optimization - build mapByType Maps as needed


jondavis9898:
Currently during UOW construction, new/dirty/delete/relationship maps are seeded with the types passed in.  Using the Application concept, this typically means all types supported by the domain layer.  However, when building a unit of work, in most cases, few very SObjects are modified.  During the commit phase, the entire map is inspected looking for operations which in most cases are empty for any given type.

To improve performance and reduce memory footprint, suggestion would be to remove the seeding of the maps from the constructor and add the key in registerDirty/New/Deleted methods.  The approach would be something along the lines of : 

1) Get SObject name
2) Find List Index from m_sObjectTypes (could change m_sObjectTypes to a Map&lt;String, Integer> and build in construction setting Integer to index order - or add another member variable that is the Map of typename to order).
3) Check the appropriate map (map key would become an Integer)
3) If key not found check if type is a supported type (via m_sObjectTypes).  If supported, add the key to the map, else throw UOWException
4) add record to map

Then, in commit, use the keyset to create a sorted list and iterate over it.

In our application, we have over 100 custom objects so a performance optimization such as this would make a significant impact as it would avoid iterating keys that are empty and also decrease the memory footprint required.

afawcett:
Ok, its a thought, given there should be just one commitWork in a request context, i'm considering what portion of this process really consumes compared to the overall execution of the request and the work it doing (querying, doing apex, other trigger code etc). Its always good to be a bit quantity in doing these type of optimisations, which i know is a little hard on the platform, but the Limits class does permit access to heap and cpu time, so i'd be interested in some comparisons between a base line use case with a full application class and say a stripped down version with a test application class just for that request to get a feel for the variance. As adding extra processing to the register methods overall could have a more detrimental effect. So yeah, i get the idea, just think some profiling is warranted to help give more confidence around such a change. Make sense?

jondavis9898:
Makes perfect sense and completely agree!  Based on your suggestion, I created some test cases/scenarios around the concepts.  In order to truly get a feel for the difference, I created a customized UOW that approached things the way I described in the OP.  Running tests against just larger and smaller Application classes wouldn't really give us a true sense of what the change would mean CPU/Heap/etc. wise, only what having more/less classes does to the current approach.

[...]

PR #63, opened 2015-07-18
Issue #58 - add newInstance overload to support performance optimization

jondavis9898:
See comments in Issue #58 - Because newInstance methods are not static, this overload could result in a mock having a different set of SObjectTypes than what is passed to this method.

### DIFF
--- a/fflib/src/classes/fflib_Application.cls
+++ b/fflib/src/classes/fflib_Application.cls
@@ -62,6 +62,22 @@ public class fflib_Application
 			return new fflib_SObjectUnitOfWork(m_objectTypes);
 		}
 
+		/**
+		 * Returns a new fflib_SObjectUnitOfWork configured with the 
+		 *   SObjectType list specified, returns a Mock implementation
+		 *   if set via the setMock method
+		 *
+		 * @remark If mock is set, the list of SObjectType in the mock could be different
+		 *         then the list of SObjectType specified in this method call
+		 **/
+		public fflib_ISObjectUnitOfWork newInstance(List<SObjectType> objectTypes)
+		{
+			// Mock?
+			if(m_mockUow!=null)
+				return m_mockUow;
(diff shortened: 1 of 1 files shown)


afawcett:
Thanks for this, i'll ponder the remark bit further.... trying to decide if its a biggie. 

@ImJohnMDaniel thoughts?

afawcett:
My initial thought was, as per yours, what if the mock doesn't support the same sobject types. Though  the mock itself is dumb anyway, in ApexMocks case its just recording what methods where called and returning responses its configured to do. If anything this change might warrent a means to mock the UnitOfWork factory tbh, but thats a different story. In short, i think this is mostly harmless and is well commented. Thanks!

ImJohnMDaniel:
Hey @afawcett and @jondavis9898.  Apologies for the delay in responding.  I read up on Jon's original idea and I think it has merit.  I appreciate Jon's predicament having a lot of SObjects in the app and the overhead that it is being introduced by it.

One question on this commit though.  Does it confuse the message some about the how to set the UOW's objectTypes?   i.e. fflib_Application.UnitOfWorkFactory( objectTypes ).newInstance( objectTypes );

afawcett:
Thats a good point, not thought about that. It does kind of make a mockery of using the factory a bit. Though the factory is there to provide instances based on the config, i guess this method is saying that the caller wants to leverage the ability of the factory to create instances (or use a mocked one) but ignore the initial / default object type list given to the factory on construction... sorta works?!?!

jondavis9898:
Agree with both @ImJohnMDaniel and @afawcett. This definitely is a "sorta works" solution and could lead to some initial confusion.  I think a better solution would be to refactor and make the intention clearer but to avoid the refactor and work with the pre-existing design, I think this is a fair compromise.  Thanks for accepting the PR!

PR #66, opened 2015-07-23
Update fflib_SObjectSelector.cls - Added ApexTrigger

tmowbrey:
Updated to include ApexClass as not needing CurrencyIsoCode


afawcett:
Thanks! :+1:

PR #68, opened 2015-07-25
Issue #56 - Support derivation of Error related classes

jondavis9898:
Decided it would be helpful in my case to also open up the ErrorFactory.  If you aren't comfortable with that additional change, let me know and I'll remove that part from the PR.  Thanks!

### DIFF
--- a/fflib/src/classes/fflib_SObjectDomain.cls
+++ b/fflib/src/classes/fflib_SObjectDomain.cls
@@ -453,16 +453,16 @@ public virtual with sharing class fflib_SObjectDomain
 	/**
 	 * Ensures logging of errors in the Domain context for later assertions in tests
 	 **/
-	public class ErrorFactory
+	public virtual class ErrorFactory
 	{
 		private List<Error> errorList = new List<Error>(); 
 		
-		private ErrorFactory()
+		public ErrorFactory()
 		{
 			
 		}
 		
-		public String error(String message, SObject record)
+		public virtual String error(String message, SObject record)
 		{
 			return error(null, message, record);	
 		}
@@ -477,7 +477,7 @@ public virtual with sharing class fflib_SObjectDomain
 			return message;	
 		}
 		
-		public String error(String message, SObject record, SObjectField field)
+		public virtual String error(String message, SObject record, SObjectField field)
 		{
 			return error(null, message, record, field);	
 		}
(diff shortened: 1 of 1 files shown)


afawcett:
Is the additional change marking all methods as 'virtual'? If so, yeah as per our earlier discussion i'd rather this be use case driven. If you can elaborate or remind me here of your use case, it might help me better understand and support this change.

jondavis9898:
Yes, additional change that wasn't originally discussed was making ErrorFactory class virtual as well as the ErrorFactory methods virtual and constructor public.

The original change discussed was confined to just making FieldError class virtual and the constructor for FieldError and ObjectError public.

The use case for making ErrorFactory is as follows...

Instead of calling addError when certain situations fail, we are adding "validation" errors to a static variable in our derived SObjectDomain class called "ValidationErrors." ValidationErrors behaves just like Errors on fflib_SObjectDomain except it's a Map<Id, List<Error>> where Id is the SObject Id.  By deriving from ErrorFactory, we can add a validation error and an error at the same time by overriding the ErrorFactory methods.  Essentially, ValidationErrors becomes a shallow reference to an error stored in ErrorFactory.  Without being able to derived, we have to add the error in both places which increases memory footprint, etc.

The error factory itself is quite useful and I can see other uses for it beyond just SObjectDomain errors.  Outside of the above though, I don't have any other specific uses cases.

If you'd prefer to not have the changes to ErrorFactory made and just have the changes to FieldError and ObjectError, no worries, just let me know and I'll remove the ErrorFactory changes from the PR.  Thanks!

afawcett:
"By deriving from ErrorFactory, we can leverage all the functionality that class provides maintaining just a shallow reference to the error in the derived class"

Would you be able to include your ValidationErrors class here so i can better understand this. As it feels like your really only leveraging its method signatures and not any of its implementation. Due to the fact you need an underlying Map and not a list of errors?

jondavis9898:
Sorry about this Andrew, you are spot on with what the net result would be.  In putting together a sample ValidationErrors for what I was thinking, I realized the ErrorFactory methods return Strings and not Error instances so my original shallow copy idea wouldn't work as I was thinking it would when I original submitted the PR.  If the method returned an Error instead of a String, I was thinking I would store off a shallow reference but defer creation of the error objects themselves to the base class.  Apologize for not looking at this more closely prior to proposing that part of the change.  

I've updated the PR to be focused on its original intent which was to simply support derivation of FieldError and ObjectError

afawcett:
Ha no worries, good to chat about these things. :+1:

PR #69, opened 2015-07-25
Issue #56 - virtualize handle methods

jondavis9898:
Debated on where in the base handleBeforeInsert/Update methods to call the new onBeforeValidation.  SFDC runs validations against layout, etc. before trigger and then again after before trigger (I think this is the way it works based on my reading).  There is something to be said for having onBeforeValidation occur as the first thing in onBeforeIns/Upd, something to be said for having it occur last and something to be said for having it occur first and last.  

For the PR, I decided to keep it simple and initiate as the first action.  With the handleXXX methods now marked virtual, the order can be customized by derived class if necessary.

Interested in your thoughts on where the before validation should fire in the sequence.

### DIFF
--- a/fflib/src/classes/fflib_SObjectDomain.cls
+++ b/fflib/src/classes/fflib_SObjectDomain.cls
@@ -159,7 +159,7 @@ public virtual with sharing class fflib_SObjectDomain
     }
     
     /**
-     * Base handler for the Apex Trigger event Before Update, calls the onBeforeValidate followed by the onBeforeUpdate method
+     * Base handler for the Apex Trigger event Before Update, calls the onBeforeValidate(Map<Id,SObject>) followed by the onBeforeUpdate method
      **/
     public virtual void handleBeforeUpdate(Map<Id,SObject> existingRecords) 
     {
--- a/fflib/src/classes/fflib_SObjectDomain.cls
+++ b/fflib/src/classes/fflib_SObjectDomain.cls
@@ -94,6 +94,16 @@ public virtual with sharing class fflib_SObjectDomain
 	}
 	
 	/**
+	 * Override this to apply general validation to be performed before insert, called by the handleBeforeInsert method
+	 **/	
+	public virtual void onBeforeValidate() { }
+
+	/**
+	 * Override this to apply general validation to be performed before update, called by handleBeforeUpdate method
+	 **/
+	public virtual void onBeforeValidate(Map<Id,SObject> existingRecords) { }
+
+	/**
 	 * Override this to apply defaults to the records, this is called by the handleBeforeInsert method
 	 **/
 	public virtual void onApplyDefaults() { }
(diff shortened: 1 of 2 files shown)


afawcett:
Yeah i must confess i thought this PR was simply addressing the situation where some of the handler methods had been marked virtual and others not. It seems you've added a new event, onBeforeValidation as well. 

The reason onValidate methods are called in the after phases, is all changes have been completed on the record, including those made potentially by other triggers, such as those added in subscriber orgs after package install, see here for [more of my thoughts on this](http://andyinthecloud.com/2015/04/19/considerations-placing-validation-code-in-an-apex-triggers/). So really its not safe to do any validation in the before. What was your motivation for adding onBeforeValidation?

jondavis9898:
Completely agree that data validation should occur in after for all the reasons you mentioned in the post you referenced.  Ironically, I initially wanted to call the method that I added "onValidateInput" rather than "onBeforeValidate" to avoid confusing it's intent.  My thinking with "onValidateInput" is that it gives an opportunity to perform validations (e.g. required fields, etc.) up-front similar to the way SFDC platform validates against page layout, etc. before the onBefore events even fire.

My use case: We have fields that are conditionally required based on data conditions that cannot be evaluated in SFDC declarative validation rules.  By having a "onValidateInput" (called "onBeforeValidate" in the PR), we have an opportunity to check these rules and fail as soon as possible rather than have the entire execution process fire only to find out they forgot the "Name" field.  In short, data validation should always be in after phases but data input can be evaluate in before to improve overall system performance, etc.  The onValidateInput provides a framework mechanism to do that and separates its concern from the normal onBeforeInsert/Update methods.

If I haven't convinced you :), just let me know and I can remove that part from the PR, no worries.  

Thanks!

afawcett:
Yeah i recall some place in the SF docs saying validation should be done in the onBefore phases of the triggers, but actually could not find it of late when writing my recently blog. As i said at the bottom of my blog on balance we optimise for the most common path, and this is hopefully valid records. Having this override in the framework i think confuses this and whats more may impact the performance valid records, since you have to repeat the validation regardless in the after phase (just in case another trigger has messed with your fields between your on before validation). So yeah, happy to be convinced otherwise in the future, but for now if you could just stick to making the remidner of the methods virtual as we discussed, i'd be happy to approve this PR.

jondavis9898:
I agree that having this in place is ripe to cause confusion.  I should have realized that when I myself struggled with coming up for a name for the method knowing that it could lead to mis-use and ultimately not following best practices.  In our case, we wouldn't be repeating the checks in the onBeforeInput in the after phase but it's definitely a situation that's outside the norm.

Short story is that I've removed the onBeforeValidate concept and only the virtualization of the handle methods is now included.  I've also renamed the PR to reflect that change.

Thanks!

afawcett:
:+1:

PR #70, opened 2015-07-25
Issue #56 - refactor permission checks

jondavis9898:
Separate permission check concerns and allow for override by derived class

### DIFF
--- a/fflib/src/classes/fflib_SObjectDomain.cls
+++ b/fflib/src/classes/fflib_SObjectDomain.cls
@@ -191,7 +191,7 @@ public virtual with sharing class fflib_SObjectDomain
     }
 
     /**
-     * Base handler for the Apex Trigger event After Insert, checks object security and calls the onValidate and onAfterInsert methods
+     * Base handler for the Apex Trigger event After Insert, calls onCheckInsertPermission then calls the onValidate and onAfterInsert methods
      *
      * @throws DomainException if the current user context is not able to create records
      **/
@@ -203,7 +203,7 @@ public virtual with sharing class fflib_SObjectDomain
     }
     
     /**
-     * Base handler for the Apex Trigger event After Update, checks object security and calls the onValidate, onValidate(Map<Id,SObject>) and onAfterUpdate methods
+     * Base handler for the Apex Trigger event After Update, calls onCheckUpdatePermission then calls the onValidate, onValidate(Map<Id,SObject>) and onAfterUpdate methods
      *
      * @throws DomainException if the current user context is not able to update records
      **/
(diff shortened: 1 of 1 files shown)


afawcett:
Thanks for this. Can you just give me a use case or two for why one would want to override these methods? The behaviour and need is pretty platform driven and there is a means to disable it, what else would you want to do in your overrides? Sorry to be a bit of a stickler for this, just make sure that if in the future we look back on some of this flexibility and ask ourselves why we have a good reason. :+1:

jondavis9898:
No worries on the questions Andrew, actually encourage that you ask them.  The conversation ultimately leads to a better product whether the ideas are included or not included.

For some background, see the comment I just made in PR #71.  In short, we can't rely on standard SFDC security and have to custom roll our own checks.  

My thinking with this PR is to separate the "permission check" concern from the handle methods.  It's similar to the way onApplyDefaults is separated in the before insert handler.  It would be easy enough to set defaults in the handleBeforeInsert but by separating it out to its own concern, it makes the code much easier to read, maintain and test.  For my situation, I could have the base class check security based on EnforceCRUD setting and then in my onBeforeXXX handlers, check custom security.  Similar to onApplyDefaults, separating the concern provides a discrete location to perform security checking.  I'm viewing security checking (like default setting, etc.) as an underlying core concept that always need to be evaluated (unless disabled) so that is the basis for this PR.  In short, just separating out the concern.

Also, as fflib looks to be enhanced through PR #71, this gives the SObjectDomain an opportunity to have ISecurity injected and defer to the injected instance rather than directly check isCreatable, etc.

Hope that helps.  Look forward to your thoughts and whether or not you think this proposal makes sense for the base library.  Thanks!

afawcett:
Yeah the fog is clearing a bit on what your trying to do with your own custom security model. I'm torn, as of course i want to see the library be generic useful and accept as many PR's as I can, and to be honest this is a fairly non contentious change and even just makes the base class code easier to manage. Though i'm still not sure what i'd say to an audience if they asked for use cases for overriding these virtual methods, e.g. what is a good use case that fits with the platforms best practice. So keen to discuss more and see your thoughts on the main discussion thread around your custom security model idea. Please understand that this is not a no to merging this, just need to understand more, lots of folks are increasingly using this library on the platform and we need to justify changes fairly well.

P.S. Would my accepting the PR that made the handler methods virtual help you here? You could switch off the base security checking and implement your own checking in the overriden ones and still delegate to the base class ones to get the routing?

jondavis9898:
Our use case is definitely outside the norm and at the end of the day, deviates from best practices.  The custom security model that we have implemented is required based on our business requirements and the limitation of SFDC that only permits a single role & single profile.  At the end of the day, it would be better to change our requirements than implement the custom security model but unfortunately that's not an option for us.

That all said, I'm with you and that coming up with a use case that would follow platform best practices would be difficult.  I do think and agree with your comment that the proposed change in this PR is non contentious and makes the base class easier to manage.  

PR #69 (make handlers virtual) would give me the flexibility I need to have a derived base that does the custom security check.  Assuming PR #69 is approved and given that from a best practice perspective there really shouldn't be a need to ever override security checking, I think we should just close out this PR.  I'll add a comment to #71 to revisit the "checkXXXPermission" refactor as that is explored.

Make sense?

afawcett:
@jondavis9898 just doing some house keeping, i see we merged PR #69, so we should be good here. Let me know if not. Thanks again for the great submissions!

PR #76, opened 2015-08-05
Rollup FF internal updates to QueryFactory

capeterson:
Deprecates creating `fflib_QueryFactory` subselects by related SObject's type. Fragile as it's non-deterministic once you have more than one relationship to that type.

Replaces it with new subselect creation methods that are safe as you extend your schema.
Adds the ability to do a deep clone of a `fflib_QueryFactory` instance.


capeterson:
Going to leave this open for review in case there are any comments. If nobody has any issues I'll merge it this weekend.

PR #77, opened 2015-08-12
Added base class for TestDataBuilders

up2go-rsoesemann:
_As promised in my [Stackexchange question](http://salesforce.stackexchange.com/questions/77387/how-to-best-use-the-test-data-builder-pattern-in-a-database-context) I hereby (ask to) contribute a new Pattern called TestDataBuilder. I think it could be of high use to many library users that have complex data structures that they need to run Apex Integration tests on. I have submitted a pull request to [fflib-apex-common](https://github.com/financialforcedev/fflib-apex-common) includung a TestDataBuilder base class and another pull request to [fflib-apex-common-samplecode](https://github.com/financialforcedev/fflib-apex-common-samplecode) with sample code showing how to use it._

There is awesome support in the library for Apex unit test. They should be fast and isolated. With [ApexMocks](https://github.com/financialforcedev/fflib-apex-mocks) [both is achieved](http://andyinthecloud.com/2015/03/22/unit-testing-with-apex-enterprise-patterns-and-apexmocks-part-1/). But complex code also needs integration tests where code is execute on more complex test data. In the world of Enterprise software outside of Salesforce.com there are experts that have created patterns for flexible and readable (fluent, concise) test data generation.

Among them the most notable is [Nat Pryce](http://www.natpryce.com/) who wrote a [great book](http://www.amazon.com/Growing-Object-Oriented-Software-Guided-Tests/dp/0321503627) about testing and somewhat invented the [TestDataBuilder pattern](http://www.natpryce.com/articles/000714.html). 

Up2Go as a Salesforce.com ISV has products with huge amount of Integration test where each of them needs to build up huge "trees" of related records. Over the year we suffered from our complex and mostly redundant TestHelpers classes where each created the same object slightly different. Whenever we changed the datamodel we had to sync all the Helpers or otherwise tests would magically fail.
- By incorporating **a single short** Builder class for each test-relevant Domain SObject we could centralize all the creation knowledge and eliminating redundancy. 
- By internally leveraging the 'fflib_SObjectUnitOfWork' for the DML all **test run dramatically faster**.
- The [Fluent API](https://en.wikipedia.org/wiki/Fluent_interface) style of the Builder pattern combined with having all the database wiring encapsulated in the Unit of work made each test much more understandable.

### DIFF
--- a/fflib/src/classes/fflib_DomainObjectBuilder.cls
+++ b/fflib/src/classes/fflib_DomainObjectBuilder.cls
@@ -1,0 +1,106 @@
+public virtual class fflib_DomainObjectBuilder {
+    
+    public SObject record;
+
+    private Map<Schema.SObjectField, fflib_DomainObjectBuilder> parentByRelationship = new Map<Schema.SObjectField, fflib_DomainObjectBuilder>();
+    
+    private static fflib_SObjectUnitOfWork uow = initUnitOfWork();
+    private static Set<fflib_DomainObjectBuilder> objects = new Set<fflib_DomainObjectBuilder>();
+    
+    
+    // CONSTRUCTORS
+
+    public fflib_DomainObjectBuilder(SObjectType type) {
+        this.record = type.newSObject(null, true);
+        
+        objects.add(this);
+    }
+    
+    
+    public fflib_DomainObjectBuilder(SObjectType type, Schema.SObjectField parentRelationship, fflib_DomainObjectBuilder parent) {
+        this(type);
+        
+        setParent(parentRelationship, parent);
+    }
+    
+    
+    public fflib_DomainObjectBuilder(SObjectType type,
+                        Schema.SObjectField leftRelationship, fflib_DomainObjectBuilder left,
+                        Schema.SObjectField rightRelationship, fflib_DomainObjectBuilder right) {
+        this(type);
+        
+        setParent(leftRelationship, left);
+        setParent(rightRelationship, right);
+    }
+    
+    
+    // PUBLIC METHODS
+    
+    public SObject build() {
+        for(fflib_DomainObjectBuilder obj: objects) {
+            uow.registerNew(obj.record);
+            
+            for(Schema.SObjectField rel: obj.parentByRelationship.keySet()) {
+                fflib_DomainObjectBuilder parent = obj.parentByRelationship.get(rel);
+                uow.registerRelationship(obj.record, rel, parent.record);
+            }
+        }
+
+        uow.commitWork();
+        
+        uow = initUnitOfWork();
+        objects.clear();
+        
+        return record;
+    }
+    
+    
+    // PROTECTED METHODS
+    
+    protected fflib_DomainObjectBuilder setParent(Schema.SObjectField parentRelationship, fflib_DomainObjectBuilder parent) {
+        
+        // Note: The parent registered last always wins!
+        fflib_DomainObjectBuilder oldParent = parentByRelationship.get(parentRelationship);
+        
+        if(oldParent != null) {
+            oldParent.clearTree();
+        }
+        
+        parentByRelationship.put(parentRelationship, parent);
+        
+        // Note: Return parent instead of this as we call this always from the parent
+        return parent;
+    }
+    
+    
+    protected fflib_DomainObjectBuilder set(String fieldName, Object value) {
+        record.put(fieldName, value);
+        return this;
+    }
+    
+    
+    protected fflib_DomainObjectBuilder set(Schema.SObjectField field, Object value) {
+        record.put(field, value);
+        return this;
+    }
+
+    
+    // PRIVATE METHODS
+    
+    private void clearTree() {
+        for(fflib_DomainObjectBuilder obj : parentByRelationship.values()) {
+            obj.clearTree();
+        }
+        
+        objects.remove(this);
+    }
+    
+    
+    private static fflib_SObjectUnitOfWork initUnitOfWork() {
+        return new fflib_SObjectUnitOfWork(new List<SObjectType> { 
+                                                    Opportunity.SObjectType,
+                                                    Product2.SObjectType,
+                                                    PriceBookEntry.SObjectType,
+                                                    OpportunityLineItem.SObjectType });
+    }
+}
--- a/fflib/src/classes/fflib_DomainObjectBuilder.cls
+++ b/fflib/src/classes/fflib_DomainObjectBuilder.cls
@@ -1,45 +1,70 @@
-public virtual class fflib_DomainObjectBuilder {
-    
+/**
+ * This class ports the TestDataBuilder pattern (http://www.natpryce.com/articles/000714.html) invented by Nat Pryce (http://www.natpryce.com/)
+ * to the world of Apex and Salesforce.com.
+ *
+ * While test data in unit tests should be as simple as possible and could even be mocked with frameworks like ApexMock,
+ * complex enterprise software also needs integration tests.The TestDataBuilder pattern reduces complexity,
+ * eliminates redundancy and improves readability while setting up complex test data structures for such tests.
+ *
+ * Learn how to replace your redundant, error-prone and test-specific helper classes with a set of small domain classes
+ * and how to write super-readable test by checking the sample code in fflib-apex-common-samplecode/blob/master/fflib-sample-code/src/classes/InvoicingServiceTest.cls
+ */
+ public virtual class fflib_DomainObjectBuilder {
+
     public SObject record;
 
     private Map<Schema.SObjectField, fflib_DomainObjectBuilder> parentByRelationship = new Map<Schema.SObjectField, fflib_DomainObjectBuilder>();
-    
+
     private static fflib_SObjectUnitOfWork uow = initUnitOfWork();
     private static Set<fflib_DomainObjectBuilder> objects = new Set<fflib_DomainObjectBuilder>();
-    
-    
+
+
     // CONSTRUCTORS
 
+    /**
+     * Constructs a standalone Domain object
+     */
     public fflib_DomainObjectBuilder(SObjectType type) {
         this.record = type.newSObject(null, true);
-        
+
         objects.add(this);
     }
-    
-    
+
+    /**
+     * Constructs a child/dependent Domain object
+     */
     public fflib_DomainObjectBuilder(SObjectType type, Schema.SObjectField parentRelationship, fflib_DomainObjectBuilder parent) {
         this(type);
-        
+
         setParent(parentRelationship, parent);
     }
-    
-    
+
+    /**
+     * Constructs a linker Domain object
+     */
     public fflib_DomainObjectBuilder(SObjectType type,
                         Schema.SObjectField leftRelationship, fflib_DomainObjectBuilder left,
                         Schema.SObjectField rightRelationship, fflib_DomainObjectBuilder right) {
         this(type);
-        
+
         setParent(leftRelationship, left);
         setParent(rightRelationship, right);
     }
-    
-    
+
+
     // PUBLIC METHODS
-    
+
+    /**
+     * Build method as defined by the TestDataBuilder pattern. Call this method on
+     * any of the Domain objects created in your test method and everything created
+     * in memory will be inserted.
+     * The later you call it the fewer DMLs your test does consume. As single call is enough
+     * but you can also call it multiple times to simulate multiple DML operations in your tests.
+     */
     public SObject build() {
         for(fflib_DomainObjectBuilder obj: objects) {
             uow.registerNew(obj.record);
-            
+
             for(Schema.SObjectField rel: obj.parentByRelationship.keySet()) {
                 fflib_DomainObjectBuilder parent = obj.parentByRelationship.get(rel);
                 uow.registerRelationship(obj.record, rel, parent.record);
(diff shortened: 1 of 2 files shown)


afawcett:
@up2go-rsoesemann really sorry for not getting to this great work sooner! DF prep is taking its usual toll this time of year on work and personal time. Anyway enough with excuses.... 

This is a really nice and light helper class and feels like a good addition. I've got some comments on the helper classes it requires but will add those to the other PR. For now this is looking good, please can i ask though that you add some more comments to the top of the class (something like your bullets and a bit of background, links etc to the pattern) and to public methods. Maybe also a short inlined class example in the class comments block?

afawcett:
Also is it possible to add @IsTest for this class?

daveespo:
Playing devil's advocate here -- Why is this pattern only useful in test contexts? We've experimented with various O/R mapping techniques to try to accomplish the same thing in our production code. That is, adding some encapsulation of bare metal SObject manipulation behind a POJO (or POAO in this case). What this pull request provides is more elegant (mostly because of the fluent pattern) than what we had and could be useful for simplifying code in Service methods

The major downside is the immense amount of Apex code that needs to be generated in order to use this with a large schema. Agreed with Andy's comment on the https://github.com/financialforcedev/fflib-apex-common-samplecode/pull/3 about wrapping the implementation classe into an outer class so your project isn't littered with dozens or hundreds of top level unit test helper classes

Lastly, is it intended that initUnitOfWork() gets modified by hand in each org/package? Why not make it virtual and protected so that a local subclass of DomainObjectBuilder can implement its own list of SObjects relevant to that org/package?

up2go-rsoesemann:
@afawcett
1)  I will add a comment to this class pointing to the sample class in the sample repository. Would that be sufficient? I think thats better and more detailed than code comment examples. Ok?! Do I need to generate another pull request or how does this work?

2) When you add @isTest to this class you also have to add it to all the implementing classes. As they are already perfectly covered (they just exist to make tests run) I see no big use in doing that. Especially as I then am NOT able to use them outside of tests as @daveespo mentions in his comment.

up2go-rsoesemann:
@daveespo 

1) Indeed this could be used outside of tests as a kind of internal API to the packages domain objects. We indeed thought about this ourselves. Even in the for of global interfaces to allow the creation of test data of base package objects in the tests of an extension package.

2) I love and will incorporate your idea to ask the implementor for a specific UnitOfWork implementation instead of modifiying the existing one. How can I modify and improve an existing Github pull request without loosing our discussion here?

daveespo:
If you commit code to your branch and reference this pull request ("#77") in your commit message, it will automatically get updated

up2go-rsoesemann:
@daveespo while trying to improve the UnitOfWork usage I come to the understanding that users of fflib always adapt the library's own UnitOfWork class. So why should they not do the same when using my (Test)databuilders?

Currently users use my pattern not by implementing a single but many classes that inherit from the abstract class. So which of them should define the UnitOfWork. Switching to what you recommentd would mean to make it more complicated to the user as besides n TestBuilder implementations he would need to provide an extra config-like class.

I'm not sure if that would add any benefit. Especially as users will have to adjust the UnitOfWork class anyway. @afawcett what's your opinion on that?

daveespo:
@up2go-rsoesemann  Hmm, I'm not entirely following what you mean by 'users of fflib adapt the library's own UnitOfWork class'?

We have created subclasses of the fflib_SObjectSelector and fflib_SObjectDomain that contain some common methods that most of our Selectors and Domain classes need -- but we haven't touched the core fflib_XX classes. We don't want to have to tediously merge in upstream changes; instead, we can just deploy the latest version of the fflib classes without modification into our source tree.

As for what I'm envisioning for your test classes:

````
public virtual class fflib_DomainObjectBuilder {
 ....
  //virtual .. or perhaps abstract?
  virtual fflib_SObjectUnitOfWork initUnitOfWork(){
    throw SomeException("You must implement initUnitOfWork() in your subclass")
  }
}

//Then a local subclass that's relevant to the scope of SObjects that you're going to be implementing in your org
public virtual class MyObjectBuilder extends fflib_DomainObjectBuilder {
  fflib_SObjectUnitOfWork initUnitOfWork(){
    return new fflib_SObjectUnitOfWork(new List<SObjectType> {Opportunity.SObjectType, Contact.SObjectType, OpportunityContactRole.SObjectType});
  }
}

//Then each implementation subclasses from MyObjectBuilder
public class Opportunity_t extends MyObjectBuilder{
  ...
}
````

I realize that initUnitOfWork is currently declared as static .. but is there a good reason for that? And it doesn't seem like there's a need to instantiate it as part of the static initializer for fflib_DomainObjectBuilder. You could just construct it within the build() method as a local variable that you let get disposed when the method exits.

I'm having a little trouble wrapping my mind around the build() method .. it doesn't feel like it belongs in fflib_DomainObjectBuilder since an instance of fflib_DomainObjectBuilder is representing a single record, yet build() operates on all records of all types. fflib_DomainObjectBuilder is acting as both a collection and a single instance.  But perhaps I'm overthinking that one ..

up2go-rsoesemann:
@daveespo I understand your confusion when looking at the build method. Let me try to explain why it is as it is.

`build()` does more than inserting on a single domain object. It will bulk-insert (to save DML and care about relationships under the hood) on every DomainObject which is not yet inserted in that (=execution context). Therefore it indeed behaves like a static singleton method. 

To make the code more readable and not introduce a hard to understand class method I gave it the object method look it has.

up2go-rsoesemann:
@daveespo I understand your confusion when looking at the build method. Let me try to explain why it is as it is.

`build()` does more than inserting on a single domain object. It will bulk-insert (to save DML and care about relationships under the hood) on every DomainObject which is not yet inserted in that (=execution context). Therefore it indeed behaves like a static singleton method. 

To make the code more readable and not introduce a hard to understand class method I gave it the object method look it has.

To come back to the original question..there is no single point of entry where a test developer could pass in a UnitOfWork. He eighter has to do it in each test (BAD!) or one time per org, by adapting the UnitOfWork class itself.

daveespo:
My point of the "MyObjectBuilder" example above is that by making initUnitOfWork virtual in fflib_DomainObjectBuilder, you don't have to modify fflib_DomainObjectBuilder.cls in every source tree to add the list of SObjects you manipulate. It remains pristine and can track upstream without any messy merges. Are we talking past each other on that point?

As for build() -- I get it -- it's just that it's not really OO to provide an instance method that operates on way more than an instance. I don't think I want to belabor that point though .. the convenience seems to outweigh the religion in this case

jondavis9898:
Great discussion everyone!  Having something in the base fflib for builder pattern makes a ton of sense.  A few thoughts:

1) The current implementation builds SObjects with DML.  When it comes to unit tests and mocking, often times the desire is to avoid DML.  Adding in a mechanism to build with or without DML would make the class significantly more valueable.

2) Agree with @daveespo regarding moving the call to "initUnitOfWork" to the build method itself and to make it virtual.  Similar to what he has done, we have derived classes on the core fflib classes (e.g. Domain) which the "real" classes then derive from.  This avoids having to change core fflib code.  When it comes to the builder class, we would have a intermediate class that overrides initUnitOfWork and then all of our SObject specific builders would derive from our intermediate class.

3) Regarding the build method itself, the builder pattern is typically patterned against an single instance and building anything necessary for that instance so having build be instance level seems appropriate to me.  With that said, in order to reduce DML as much as possible, possibly build is instance level and persist is static.  This helps with providing a way to create SObjects using DML (lots of stuff via persist) and not using DML (single SObject and its dependencies via build).  

I'm currently working on implementing the pattern with the modifications described above.  I'll share to @up2go-rsoesemann repo once I have something consumable.

up2go-rsoesemann:
@jondavis9898 glad that you see a need for (Test)Builder support in this library. I am more than happy to see you improve some of the glitches in the current implementation. I am looking forward to hear and see more.

jondavis9898:
@up2go-rsoesemann  I submitted PR #100 with my suggestions on enhancements/extensions.  Thank you again for getting the ball rolling on this, look forward to yours (and others) thoughts!

afawcett:
@up2go-rsoesemann Just wanted to drop in and say that i've not forgotten about this, and am pleased to see others chipping in on including this.

afawcett:
@jondavis9898 great work on the PR https://github.com/financialforcedev/fflib-apex-common/pull/100. @up2go-rsoesemann what do you think?

I'm keen to get this live actually, it seems folks see a need for it and although i still share concerns over the amount of code it generates. At the end of the day that is up to the developer and if they see the value then the over head is worth it. And maybe one day we can get Salesforce to generate somethign like this, the way they have now provided support in the platform that avoids the need for generating Apex Mocks mock classes! 👍 

So next steps....
- @jondavis9898 and @up2go-rsoesemann to agree Jon's PR is good
- Given the motivations and discussions here https://github.com/financialforcedev/fflib-apex-common/issues/36 i think its worth creating a plug-in / extension repo that can be layered on optionally to this core library. Something fflib-apex-common-builder ? Would this be ok with everyone? If so, i'll create the repo and Jon you could generate a PR on that.

How this work for everyone?

up2go-rsoesemann:
Thanks @afawcett for giving this a second chance. We regard our internal Builder architecture as a huge success. Our tests are way more readable, faster and nearly no redundancy in test setups.

What Jon has added is HUGE. Thanks for you awesome work. I will need some time to dig into his changes. jondavis9898 is there a TL:DR fast lane to sample code that shows how the builder is used? Maybe a test class that shows all the great new features in action?

afawcett:
Excellent, thanks @up2go-rsoesemann. @jondavis9898 your thoughts?

up2go-rsoesemann:
@jondavis9898 I could answer all of my questions by re-reading all the PRs and comments. Maybe I have a simplicity/readability fetish, but I still think we should do some minor tweaks to the API:

1. Object Mother pattern is a great idea. But I also want to use a simple constructor:
`new Opportunity_t().name('...')` instead of `Opportunity_t. anOpportunity().name('...')`
2. Having the line `fflib_DomainObjectBuilder.persistRegistered(Application.UnitOfWork.newInstance());` in all my test methods is ugly. Can't we instead have a `fflib_DomainObjectBuilder.persist()` and define the UnitOfWork somewhere else? Maybe in a static block using `fflib_DomainObjectBuilder.use(UnitOfWork...)`?!
3. Please keep the library independent from all the fflib interfaces required for ApexMocks. I saw you using the Application class. We don't use them and many other (those that ask for a lighter fflib) don't eighter.

jondavis9898:
Hello guys - Glad to see this re-surfacing!

Agreed that this should be a separate repo.

@up2go-rsoesemann Addressing your 3 questions/comments:

1) PR #100 requires that you use "new" to create an instance.  It does not require nor prohibit the use of Object Mother.  Using the Object Mother pattern in a derived builder class is optional.

2 part a) The reasoning behind naming persistRegistered is explained in more detail in the comments to PR #100 ([here](https://github.com/financialforcedev/fflib-apex-common/pull/100#issue-128099838) and  [here](https://github.com/financialforcedev/fflib-apex-common/pull/100#issuecomment-174109152)).  In short, derived builders can expose persist() so that they can return a concrete SObject.  If the base class offered persist(), then derived builders would need another name for "persisting" an instance (and it's related builders).  The base class provides complete flexibility on usage but the most common use case within an actual test is to call the instance level persist() (which persists everything that has been registered).  You can still use persistRegistered static from the base if you choose of course.  Given this flexibility, my goal was to have the names be as similar as possible - hence, persistRegistered(UnitOfWork)/persistBuilder(UnitOfWork)/persist(UnitOfWork, Set<builders>) on the base class allowing the derived class to offer persist().  See [here](https://github.com/financialforcedev/fflib-apex-common-samplecode/pull/6/files#diff-b655d108aa2ce4309c81bd5ba11257beR160) for an example.

2 part b) Given the discussion with @daveespo above, my thinking was to move to more instance level methods than statics to provide flexibility, specifically around being able to create builder instances with or without DML (to your question earlier, this is one of the primary differences from your original PR #77 which only provided DML based option).  Additionally, the approach allows for flexibility around which UnitOfWork to use as well as providing default UOWs within a derived builder class instead of having the test itself have to create one.  Having a "use" to me is ambiguous because when is it "used"?  Keeping with the "persist" and "build", the intent/purpose is clearer I believe.

3) PR #100 has no dependency on ApexMocks or using the Application class concept from fflib.  Please let me know where you see this dependency, possibly I'm overlooking it?

Not sure if you saw but I did submit a [PR to MavensMate templates](https://github.com/joeferraro/MavensMate-Templates/pull/18).  If you MM, creating builders is extremely simple.  If you don't use MM, its still rather simple by following the structure outlined in the template or fflib sample repo tests.

Look forward to your thoughts/feedback!

up2go-rsoesemann:
@jondavis9898 Sounds more than ok to me. Thanks for detailing this again. 

@afawcett Please go on with making this public as fflib-apex-common-builder.

jondavis9898:
👍

afawcett:
Done! https://github.com/financialforcedev/fflib-apex-common-builder

up2go-rsoesemann:
Cool.

@jondavis9898 would you mind writing a short Readme.md with a few use cases?
@afawcett will you add the code to the repo? And how should we contribute to it? Via forking or can you make @jondavis9898 and me a collaborator (please also add my collegue @up2go-sschwarz)?

afawcett:
@up2go-rsoesemann Please feel free to produce a PR to start to populate the repo.

jondavis9898:
Hey guys - Awesome seeing this getting so close!  I'll submit a PR with the code as soon as I can.

@up2go-rsoesemann This was your idea initially so I'll defer to you on the readme, use cases, etc.  Happy to collaborate once you get the initial version going!

up2go-rsoesemann:
@jondavis9898 would you mind getting your code in first so I can play around with it to add a few first code examples done?

jondavis9898:
Yep, was thinking you'd want the code first, just been swamped and haven't had time to get it uploaded yet.  Will do as soon as I can, sorry for the delay!

up2go-rsoesemann:
Ok @jondavis9898 I created a PR with my old classes so we have a history. I will also add an initial README soon.

jondavis9898:
@up2go-rsoesemann  @afawcett  Sorry for the delay on getting the PR submitted on this!  I just submitted https://github.com/financialforcedev/fflib-apex-common-builder/pull/2 to the new repo.

@up2go-rsoesemann I updated the README that @afawcett added with some pointers to the sample code and MM templates.  As you update the readme, please be sure to include references to those.  Thanks!

afawcett:
Thanks @jondavis9898

Issue #82, opened 2015-09-07
Create Domain Class from non-concrete list of SObjects


jondavis9898:
Currently a concrete list of SObjects is required in order to create an instance of an fflib_SObjectDomain class.  The reason for this is that the constructor of fflib_SObjectDomain calls [getSObjectType()](https://github.com/financialforcedev/fflib-apex-common/blob/master/fflib/src/classes/fflib_SObjectDomain.cls#L91) on the list specified in the constructor arguments and this call will return null if the List is not of a concrete type (e.g. Opportunity, Account, etc.).

Use Case: In an approach similar to what fflib UOW does, I have a need to keep a Map of SObjects by type and then iterate the map, constructing a domain class and invoking methods on it (methods are exposed via common interface).  Currently, my only solution is to keep a static map of concrete Lists to SObjectTypes which can be cloned and used within an execution context.  This carries with it additional maintenance/support overhead in managing an "Application" class as well as the perf cost incurred cloning large maps (we have 100+ domain classes).  An alternate solution would be to add a constructor to fflib_SObjectDomain allowing the SObjectType to be supplied instead of inferred from the List supplied. 

Unfortunately, there is no way to create an List of a specific SObjectType at runtime.  Given the goal of SObjectDomain classes and ability to use them outside of a trigger context (which supplies concrete lists in Trigger Maps), adding the constructor seems to make sense and provide a mechanism to do.

Interested in any feedback and especially for those more familiar with the entire fflib package, if not having a concrete list of "Records" could present any downstream impacts.

afawcett:
Thanks for the PR! :+1:  Merged.

jondavis9898:
tag this as enhancement for tracking purposes?

afawcett:
Done. :+1:

PR #83, opened 2015-09-07
Issue #82-Add support for creating SObjectDomain classes from generic lists

jondavis9898:
In order to maintain full backwards compat and not impact performance of existing library, kept the code path completely separate from existing code path using IConstructable2 interface.  This resulted in some duplication of code ([fflib_Application.DomainFactory.newInstance overload](https://github.com/jondavis9898/fflib-apex-common/commit/77aabb799ca3f81792ad57641dd3011684f75840#diff-c304b0b975b1713fc151a191808e44fcR318)) but merging the newInstance methods would have resulted in a perf hit to check for IConstructable2 and also possibly using IConstructable2 when the caller really wanted to use IConstructable (without a few more unnecessary conditional checks).  While this still functionally would have worked, the small amount of duplicate code in newInstance vs. the perf hit trade-off seemed like a fair compromise.  If refactoring and/or extending IConstructable would have been an option, there are better ways of adding this enhancement.

Note that [this line](https://github.com/jondavis9898/fflib-apex-common/commit/77aabb799ca3f81792ad57641dd3011684f75840#diff-6cd4e022dccbea4cec890b12f6d560fdR383) of code from existing codebase seems a little odd since the constructor accepts a List&lt;Opportunity> instead of List&lt;Account>.  Not sure if this is a defect or intentional?
```
public AccountsDomain(List<Opportunity> sObjectList)
{
	super(sObjectList);
}
```

### DIFF
--- a/fflib/src/classes/fflib_Application.cls
+++ b/fflib/src/classes/fflib_Application.cls
@@ -304,6 +304,37 @@ public class fflib_Application
 			return (fflib_ISObjectDomain) domainConstructor.construct(records);
 		}	
 
+		/**
+		 * Dynamically constructs an instace of the Domain class for the given records and SObjectType
+		 *   Will return a Mock implementation if one has been provided via setMock
+		 *
+		 * @param records A list records
+		 * @param domainSObjectType SObjectType for list of records
+		 * @exception Throws an exception if the SObjectType is not specified or if constructor for Domain class was not registered for the SObjectType
+		 *
+		 * @remark Will support List<SObject> but all records in the list will be assumed to be of
+		 *         the type specified in sObjectType
+		 **/
+		public fflib_ISObjectDomain newInstance(List<SObject> records, SObjectType domainSObjectType)
+		{
+			if(domainSObjectType==null)
+				throw new DeveloperException('Must specify sObjectType');
+
+			// Mock implementation?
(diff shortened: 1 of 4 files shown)


afawcett:
Looks great thanks! :+1:

jondavis9898:
Thanks @afawcett.  Any thoughts on the last comment about the odd looking construction parameter for AccountsDomain (Opportunity instead of Account)?  Seems like this should be changed to List of Account but didn't want to "fix" it without an additional set of eyes as possibly it was intentional.

afawcett:
Oh yes, that looks like an oversight. Maybe unused code?

jondavis9898:
It appears to be used, that's the strange thing as I'm not sure how it's even working.

[This line](https://github.com/financialforcedev/fflib-apex-common/blob/master/fflib/src/classes/fflib_ApplicationTest.cls#L398) of code will call the single param constructor passing in a List of Accounts from [this line](https://github.com/financialforcedev/fflib-apex-common/blob/master/fflib/src/classes/fflib_ApplicationTest.cls#L398).

Thoughts on how this could even be working?  Apex bug?

afawcett:
It does feel like an Apex bug to me, thanks for spotting it!

PR #91, opened 2015-11-03
Update RecordType.sObjectType.getDescribe().getName()

tmowbrey:
RecordType.sObjectType.getDescribe().getName() returns "SObjectType". We need to use RecordType.getSObjectType().getDescribe().getName() to return RecordType. I've found that I need to use RecordType.getSObjectType() in place of RecordType.SObjectType everywhere in this library.


afawcett:
Most interesting difference in behaviour for this object. It's not related to this library but in general i've run this in dev console, most bizzare...

    System.assertEquals('Attachment', Attachment.SObjectType.getDescribe().getName());
    System.assertEquals('ApexClass', ApexClass.SObjectType.getDescribe().getName());
    System.assertEquals('SobjectType', RecordType.SObjectType.getDescribe().getName());
    System.assertEquals('RecordType', RecordType.getSObjectType().getDescribe().getName());

PR #95, opened 2015-11-29
Key assertIsAccessible based on boolean value

tmowbrey:
It appears as though the variable passed into this method is never checked before asserting that the object is available.


afawcett:
Thanks, good spot!

Issue #36, opened 2015-03-24
Enterprise Patterns 'Light'


john-m:
Firstly - thanks for sharing the patterns!

This post was prompted by a brief exchange with Andrew after this tweet :

<blockquote class="twitter-tweet" lang="en"><p>Great call with <a href="https://twitter.com/LaceySnr">@LaceySnr</a> this morning, talking about Apex Enterprise Patterns “Light” and consulting project usage! <a href="https://twitter.com/SalesforceDevs">@SalesforceDevs</a></p>&mdash; Andrew Fawcett (@andyinthecloud) <a href="https://twitter.com/andyinthecloud/status/573760842543554560">March 6, 2015</a></blockquote>

<script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>

At the time I was trying to decide whether to use the patterns in my upcoming project but was concerned about the complexity. Hence, my interest in a ‘Light’ version…

We’ve been on Salesforce for about 18 months and our implementation, although not particularly large, covers quite a lot of the platform with two-way integrations to our website on the frontend and our legacy ERP on the backend, scheduled & batch Apex plus the usual VF controllers and triggers etc. At this stage I'm starting to feel the maintenance pain and therefore very receptive to the ‘separation of concerns’ approach.

I’m the business owner and spend only a small part of my time as a ‘developer’, going months (if all is running well) without touching the code… but we’re about to embark on another phase of implementation involving FinancialForce Accounts & HCM and then re-implementing what’s left of the ERP in Salesforce. 

In short I want maximum separation of concerns to help with maintainability but with minimum complexity!

I've seen [this post](http://www.laceysnr.com/2015/03/the-enterprise-patterns-on-diet.html) from Matt Lacey but I’m not confident enough to go pulling the framework apart and I have a feeling that I’m going to find some of the more advanced parts, eg mocks, useful once I get into it, so my approach to 'Light' so far is to just pick what seem like the more straightforward and approachable parts and start there. Something like this :
- **Application** class that just implements Unit of Work
- **Service** classes - I’d already started collecting related methods into classes so this involves the adoption of a more formal and disciplined approach to this, and Unit of Work is helpful here
- **Selector** classes - pretty straightforward to implement and address an issue that I’ve already experienced trouble with (consistency of fields returned via SOQL)
- **Domain** class - this one seems less obvious but plan to try it out next time I need to code a trigger
[...]

PR #100, opened 2016-01-22
Extend PR #77 - Add Test Data Builders

jondavis9898:
This is an extension of the concepts originally introduced in PR #77 by @up2go-rsoesemann.  A big Thank You to him for getting this started!

Enhancements from #77:
1. Add ability to build existing or new SObjects in-memory 
*\* [build(Boolean)](https://github.com/jondavis9898/fflib-apex-common/blob/pr-77-add-testdatabuilders/fflib/src/classes/fflib_DomainObjectBuilder.cls#L139) on Base / [build()](https://github.com/jondavis9898/fflib-apex-common/blob/pr-77-add-testdatabuilders/fflib/src/classes/fflib_DomainObjectBuilderTest.cls#L1074) & [buildNew()](https://github.com/jondavis9898/fflib-apex-common/blob/pr-77-add-testdatabuilders/fflib/src/classes/fflib_DomainObjectBuilderTest.cls#L1083) on derived
2. Persist single builder to DB or register multiple builders for persistence to DB
*\* Single - [persistBuilder(fflib_ISObjectUnitOfWork)](https://github.com/jondavis9898/fflib-apex-common/blob/pr-77-add-testdatabuilders/fflib/src/classes/fflib_DomainObjectBuilder.cls#L197) on base / [persist(fflib_ISObjectUnitOfWork)](https://github.com/jondavis9898/fflib-apex-common/blob/pr-77-add-testdatabuilders/fflib/src/classes/fflib_DomainObjectBuilderTest.cls#L1092) on derived
*\* Multiple - [persistRegistered(fflib_ISObjectUnitOfWork)](https://github.com/jondavis9898/fflib-apex-common/blob/pr-77-add-testdatabuilders/fflib/src/classes/fflib_DomainObjectBuilder.cls#L100) on base
3. Test Coverage - [fflib_DomainObjectBuilder](https://github.com/jondavis9898/fflib-apex-common/blob/pr-77-add-testdatabuilders/fflib/src/classes/fflib_DomainObjectBuilderTest.cls)
4. Add Object Mother pattern for simplified builder creation

Comments:
1. Per discussion in #77:
*\* Rather than have builder create Unit Of Work, decided to have it be a method parameter.  The reason for this is a couple of fold but ultimately, every test data situation is slightly different and if list of SObjectTypes is maintained in the builder itself, it limits what SObject can be built during persist operations.  By passing in as a parameter, the test setup code can ensure the UOW has the correct SObjectTypes.  Also, the derived class can choose to have a persist() and a persist(fflib_ISObjectUnitOfWork) convenience methods that both wrap persistBuilder(uow).  The persist method can create a 'default' UOW internally and when necessary, persist(uow) can be called when the default of the builder does not satisfy the needs.  In short, get the best of both worlds here I think.
*\* Persistance to the DB when operating on registered builders held in static is done via a static method
*\* Persistance to the DB on an individual builder instance (and it's related data) is done via instance method
*\* Decided not to add @IsTest to fflib_DomainObjectBuilder since this class can be used for non-test purposes (see future enhancements below as well) and has 100% test coverage.
*\* Regarding a single file per builder or one file with lots of inner builders, this is really the choice of the implementer and I don't think having to choose one way or the other as it relates to fflib itself is required.
2. Chose to be a little verbose on method names in base class (e.g. persistBuilder, registerBuilder, etc.).  While these methods are marked public, the intention is that derived classes will always wrap these methods exposing a shorter name (e.g. persist, register).  This allows the derived class to cast to its own type (or the specific SObject type) avoiding the caller from having to cast return values all over the place and also allowing for fluent configuration in certain situations (e.g. registerBuilder).
3. fflib_DomainObjectBuilderTest contains samples of how to leverage the library for all 4 "build" methods.  Review any of the TestXXXBuilder inner classes for the standard methods that would/should be included in all derived classes.
4. Liberal use of protected & virtual allowing for derived classes to extend as needed

Future Enhancements:
1. Test Data Builder generator similar to Apex Mocks Generator
2. Ability to handle 'updates' of SObject data not just inserts making the pattern more useful in production situations
3.  ~~Updates to fflib-apex-samplecode demonstrating pattern (samples already exist in fflib_DomainObjectBuilderTest)~~ - See [Samples](https://github.com/financialforcedev/fflib-apex-common-samplecode/pull/6)
4. ~~MavensMate template~~ - See [Templates](https://github.com/joeferraro/MavensMate-Templates/pull/18)

Feedback encouraged and welcome.  Look forward to everyone's thoughts!

### DIFF
--- a/fflib/src/classes/fflib_DomainObjectBuilder.cls
+++ b/fflib/src/classes/fflib_DomainObjectBuilder.cls
@@ -1,0 +1,450 @@
+/**
+ * Copyright (c) 2014, FinancialForce.com, inc
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, 
+ *   are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice, 
+ *      this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice, 
+ *      this list of conditions and the following disclaimer in the documentation 
+ *      and/or other materials provided with the distribution.
+ * - Neither the name of the FinancialForce.com, inc nor the names of its contributors 
+ *      may be used to endorse or promote products derived from this software without 
+ *      specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND 
+ *  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES 
+ *  OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL 
+ *  THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, 
+ *  EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ *  OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ *  OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+**/
+
+/**
+ * Base class aiding in the implementation of Test Data Builder pattern as described by Nate Pryce
+ * (http://www.natpryce.com/).  Derived classes can combine the features the Builder 
+ * Pattern (http://www.c2.com/cgi/wiki?BuilderPattern) with an Object Mother pattern 
+ * (http://www.c2.com/cgi/wiki?ObjectMother) to simplify the setup/creation of data.
+ * 
+ * While test data in unit tests should be as simple as possible and could even be mocked with frameworks like ApexMock,
+ * complex enterprise software also needs integration tests. The TestDataBuilder pattern reduces complexity,
+ * eliminates redundancy and improves readability while setting up complex test data structures for both
+ * unit and integration tests and can also be used for production purposes.
+ *
+ * Learn how to replace your redundant, error-prone and test-specific helper classes with a set of small domain classes
+ * and how to write more readable tests by checking the samples in the test class fflib_DomainObjectBuilderTest
+ */
+public abstract class fflib_DomainObjectBuilder 
+{
+	/**
+	 * @description Maintains non-relationship field values that have been assigned to an instance of this class
+	 **/
+    protected Map<Schema.SObjectField, Object> m_fieldValueMap = new Map<Schema.SObjectField, Object>();
+
+	/**
+	 * @description Maintains relationship field values that have been assigned to an instance of this class
+	 *              Relationship field values are references to builder instances
+	 **/    
+    protected Map<Schema.SObjectField, fflib_DomainObjectBuilder> m_parentByRelationship = new Map<Schema.SObjectField, fflib_DomainObjectBuilder>();
+
+    /**
+     * @description The SObjectType this Builder class represents
+     **/    
+    protected Schema.SObjectType m_sObjType;
+
+    /**
+     * @description The SObject this Builder class is building
+     **/    
+    protected SObject m_sObjRecord { get; protected set; }
+
+    /**
+     * @description Tracks all builders that have registered to be persisted to a UnitOfWork
+     **/    
+    private static Set<fflib_DomainObjectBuilder> m_registeredBuilders = new Set<fflib_DomainObjectBuilder>();
+
+	/**
+	 * @description Constructs the Builder class with the specified SObjectType
+	 *
+	 * @param type The SObject type that the builder will build
+	 **/
+    protected fflib_DomainObjectBuilder(SObjectType type) {
+        this.m_sObjType = type;
+        // passing false because of bug in SFDC - See https://success.salesforce.com/issues_view?id=a1p30000000Sz5RAAS
+        // not establishing defaults also gives us an opportunity for a completely "clean" record        
+        this.m_sObjRecord = m_sObjType.newSObject(null, false);
+        this.IsBuilt = false;
+        this.IsRegistered = false;
+    }
+
+	/**
+	 * @description Copy Constructor that constructs the Builder class based on the builder specified
+	 *
+	 * @param copyFrom The builder to copy/clone this instance from
+	 **/
+    protected fflib_DomainObjectBuilder(fflib_DomainObjectBuilder copyFrom)
+    {
+    	this(copyFrom.getSObjectType());
+    	m_fieldValueMap = copyFrom.m_fieldValueMap.clone();
+    	m_parentByRelationship = copyFrom.m_parentByRelationship.clone();
+    }
+
+    /**
+     * @description Commits all registered builders and their related builders to the Unit Of Work specified
+     *
+     * @param uow The Unit Of Work to process against
+     **/    
+    public static void persistRegistered(fflib_ISObjectUnitOfWork uow) {
+    	persist(uow, m_registeredBuilders);
+    }
+
+    /**
+     * @description Returns true if builder has been built, false otherwise
+     *              A builder IsBuilt build or persist operations have completed successfully on it
+     **/
+    public Boolean IsBuilt { get; protected set; }
+
+    /**
+     * @description Returns true if builder has been registered, false otherwise
+     *              A builder IsRegistered if registerBuilder has been called on it and the
+     *              builder has not yet been built/persisted
+     **/    
+    public Boolean IsRegistered { get; protected set; }
+
+    /**
+     * @description Returns the SObject associated to this builder
+     **/    
+    public SObject getRecord() {
+        return m_sObjRecord;
+    }
+
+    /**
+     * @description Returns the SObjectType associated to this builder
+     **/    
+    public virtual Schema.SObjectType getSObjectType() {
+        return m_sObjType;
+    }    
+
+    /**
+     * @description Returns the SObject after building the builder and any of its related builders
+     *
+     * @param isNew True if the instance should not have an Id value (new record), false otherwise (existing record)
+     *
+     * @remarks Recommended to wrap this method with a 'build()' and 'buildNew()' in a derived class
+     *          casting to the derived type to support fluent configuration
+     **/  
+    public virtual SObject build(Boolean isNew)
+    {
+    	// see if we'll let 'em through
+        checkAllowBuild();
+
+        // fire event
+        beforeBuild(isNew);
+
+        // set non-relationship field values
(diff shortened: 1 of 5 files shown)


up2go-rsoesemann:
@jondavis9898 thanks for giving my idea a new chance of being merged by adressing some of the problems. 

I totally get the rationale behind your changes. I am also quite sure that you implemented them thoroughly. Your changes (actually a rewrite) are way to massive to understand them without really digging in for hours ;-).

But..what I can say from looking at the tests for 5 minutes is that the resulting test don't look like I would like to have them. I don't care if I don't understand the implementation anymore (thats why I use a library). But...if the result doesn't make my test more readable and concise and tech free I won't use it. There are TestXYZBuilders and pattern-related names all over the place. I would say the implementation details shine through... ;-)

I think some renaming in your classes public interface could probably do the job. Maybe you could also rebuild my test implementation (https://github.com/up2go-rsoesemann/fflib-apex-common-samplecode/commit/adab15414782cff1ffb914f89d211c5e7b283d2e) with your implementation and see how clean and readable the are.

jondavis9898:
@up2go-rsoesemann Thanks for reviewing the proposed changes and providing your feedback!

Regarding the tests for fflib_DomainObjectBuilder itself and the TestXYZBuilder naming convention, these are intended to be tests for the Builder itself and not meant to represent what a naming convention in actual use might be.  That said, it does demonstrate that the name of the actual "builders" an be anything that the consumer of the builder desires as it doesn't influence the functionality.  For example, your example used Opportunity_t while someone else could use OpportunityBuilder, etc.  In short, apologize if the naming convention made things more challenging to understand.

I have updated the sample code repo with example using using the naming convention that you chose.  I wrote the same test that you did in there as well as several others to demonstrate all the functionality of fflib_DomainObjectBuilder.  Please have a look and let me know if this helps clarify usage.

I agree that the naming convention on the public interface of the fflib_DomainObjectBuilder is a little cumbersome.  When writing a test, other than the persistBuilders(uow) method, you would never call a method on fflib_DomainObjectBuilder directly.  Instead, you would call methods on the derived classes which, as you'll see in the sample, are shorter and easier to follow.  

The reason I chose to do it this way is two-fold:
1. As I mentioned in my previous post above, it avoids having to cast the result of the method call whenever you build/persist something within a test.  By having the derived classes expose "build()" and call fflib_DomainObjectBuilder's build(Boolean), the derived class build() can cast to the specific SObjectType it needs.  If the test code called a build method on fflib_DomainObjectBuilder directly, each call would need to be cast in the test code itself making things much more difficult to follow/read within the test itself.  
2. I've added 3 new ways of creating SObjects using the builders but wanted to keep the naming similar - this caused method names to be a little more verbose in the base class itself.  As an example, with the ability to create in-memory SObjects and not just DB based, you can use the builder to write a true unit test that doesn't require the database, something that wasn't possible in the original version of fflib_DomainObjectBuilder (see [testService_UsingBuildersWithoutDML](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/InvoicingServiceTest.cls#L107)).

Keeping with trying to make tests as easy to read/follow as possible, I've also implemented an Object Mother pattern to avoid having "new" keywords all over the tests.  

In addition to adding the sample code, I've added MavensMate templates at https://github.com/joeferraro/MavensMate-Templates/pull/18.  By creating a new DomainObjectBuilderBase and then specific DomainObjectBuilder classes, the boiler plate code really leaves you only to have to add "withXXX" methods to the build classes themselves and then write your tests.

Here's the approach:
1) Create DomainObjectBuilderBase (e.g. Domain_t)
2) Modify createUnitOfWork() on DomainObjectBuilderBase as appropriate (see comments in method from template)
3) Create DomainObjectBuilder (e.g. Account_t, Opportunity_t) for each SObject you want to create a builder for
4) Add "withXXX" methods to the builders as appropriate
5) Optional: Add additional Object Mother static methods to the builders as appropriate
6) Write your tests and use the following:
Instance Methods
[build()](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/InvoicingServiceTest.cls#L120) - Builds an in-memory Existing SObject (Id is NOT null) and returns the SObject cast to the derived SObjectType
[buildNew()](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/OpportunitiesTest.cls#L76) - Builds an in-memory New SObject (Id is null) and returns the SObject cast to the derived SObjectType
[persist()](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/OpportunitiesTest.cls#L160) - Commits the builder and any of its related data through the default Unit Of Work (obtained via createUnitOfWork) and returns an instance of the SObject cast to the derived SObjectType
[persist(uow)](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/OpportunitiesTest.cls#L202) - Commits the builder and any of its related data through the Unit Of Work specified and return an instance of the SObject cast to the derived SObjectType
[register()](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/InvoicingServiceTest.cls#L95) - Registers the builder for commit when persistBuilders(uow) is called
[Record](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/InvoicingServiceTest.cls#L100) - Returns the SObject cast to the derived SObjectType
[but()](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/OpportunitiesTest.cls#L113) - Creates a clone of the builder instance

Static Methods on Derived Builders
[Object Mother Helpers](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/OpportunitiesTest.cls#L110)

Static Methods on Base
[persistRegistered(uow)](https://github.com/jondavis9898/fflib-apex-common-samplecode/blob/pr3-add-testdatabuilders/fflib-sample-code/src/classes/InvoicingServiceTest.cls#L97) - Commits all registered builders and any of their related data through the Unit of Work specified

I'm hopeful that once you review the sample code, you'll see how all this plays out and how clean the actual test itself is.  Hope this helps, let me know what you think and any questions you might have.  Thanks!

up2go-rsoesemann:
Thanks for clarifiying you concepts. Looks like there are dependancies to ApexMocks and its related classes (e.g. Application). I never bought into applying those concepts in Apex and hope that dicussions like #36 will lead to a more lightweight direction.

jondavis9898:
@up2go-rsoesemann Thanks for reviewing the samples.  To answer your questions:

1) There is no depdency on Application within fflib_DomainObjectBuilder which is what this PR covers.  The only dependency on "Application" is in one place in the sample code and is optional.  Since the sample code contains an Application class, createUnitOfWork() uses it.  It could very easily not use it and just create an fflib_SObjectUnitOfWork(List<SObjectType>) directly.  In the [MavensMate template](https://github.com/jondavis9898/MavensMate-Templates/blob/master/ApexClass/DomainObjectBuilderBaseClass.cls#L44) you'll see a comment that indicates there are multiple options on how to write createUnitOfWork.  In fact, you don't even need to have a createUnitOfWork at all, it's simply a helper method.  The consumer could choose not to implement persist() and only expose persist(fflib_ISObjectUnitOfWork) instead in which case createUnitOfWork() wouldn't even be needed.

2) If you wanted to add a child object to a parent, you could add a method to your Account_t called "add(OpportunityBuilder)" and then handle the relationship setting in that method.  Something like:

    public Account_t add(OpportunityBuilder value) {
        value.withAccount(add);
        return this;
    }

Please keep in mind that fflib_DomainObjectBase is simply an abstract base class.  How the consumer of the library chooses to use it is up to them.  They could name their builder Opportunity_t or OpportunityBuilder, they could use a derived class like Domain_t or choose not to, they could use Application class or choose not to.  The samplecode is just one example of how the library could be used, it doesn't require that everyone use it that way.  This is one of the reasons why there is test coverage for fflib_DomainObjectBuilder, because there is no right or wrong way to use it.  It's up to the consumer of the base class to choose how to use it and up to the base class to provide the functionality to satisfy most (if not all) use cases.

up2go-rsoesemann:
@afawcett Who under which process decides if a pull request is going to be merged in, rejected or just ignored?

afawcett:
**Do we need this kind of thing?** Ok so i've been reviewing this finally, so sorry for the delay. Firstly i want to progress this for sure! I agree for certain developers that expect this type of facility on the platforms they work on its a great addition. I'm happy and honoured to support the great community around this library to lead the charge as you guys are doing.

**ApexMocks parallel?** For me this feels much like the time Paul Hardaker and I first discussed introducing ApexMocks. We worried about the bloat and who would really understand it and i'm pleased to see its really taken off. I think the same can be true here for sure!

Some more specific thoughts...

- **In or Out this Repo?** Given the thoughts already shared about the library getting large, which i also support, i'm torn by potentially compounding this and building debt for when we also get round to splitting up this library vs the fact that this is really an additional two classes.... Robert i hear what your saying, and tbh felt the same way. However as per the discussion on the https://github.com/financialforcedev/fflib-apex-common/issues/36 its also about barrier to entry for new devs that might intimidated and ignore the whole library all together and also adoption in none ISV/packaged scenarios john-m and others are talking about.
- **Builder Classes** There is a lot of boiler plate needed for the builder classes, i think this will inhibit adoption of it tbh. Again we had the same discussions around ApexMocks, hence we ended up with a generator tool, perhaps the same approach here is needed? I also think we need to provide the option to scope the builder classes or at least in the hand written samples, show them as inner classes and not pollute the top level namespace with essentially supporting classes. 

**Proposal:** I think on balance i'm going to recommend we add this fine work into a fflib-apex-testbuilder repo. As far a test builder classes go perhaps contain these in the sample app in a TestBuilders class, hand maintained for now, but then perhaps think about a generator much like ApexMocks.

jondavis9898:
Thanks for reviewing Andrew, glad to here you feel a builder concept like this has merit.

I'm in full support of a separate repo.  I think using the builder pattern is completely optional and doesn't need to be in apex common core.  Personally, I'd like to have it in the core library but in my use cases, I'm using everything in fflib so it's nice to just have one place to manage everything from.  That said, I think there is very good merit to having thinks like Mocks, Builder and a few others in separate repos since they aren't "core".

Regarding inner classes vs. separate classes, I feel that decision is completely up to the consumer of the pattern/library and the library itself doesn't need to dictate.  Demonstrating both options makes sense and provides flexibility.  In the test cases for the builder I used inner classes, however in apex sample code, I used separate classes.  Along these same lines, I feel that naming convention should also be at the consumers discretion (e.g. Account_t vs. AccountBuilder, etc.).

Regarding a generator, I actually started writing one when I updated this PR but unfortunately, I didn't get very far and finishing it has been put on the backburner do to other priorities.  My design calls for a flag that would emit the classes as inner classes or separate so I think that is in-line with your suggestion.  The design also called for options for naming pattern, what the "Object Mother" default method should be named, etc.  If someone wants to jump in and write the generator, I'd be all for that!

In the meantime, I did update [mavensmate templates](https://github.com/jondavis9898/MavensMate-Templates/blob/master/ApexClass/DomainObjectBuilderBaseClass.cls#L44) with templates for the builders (also updated Domain template since it was a little out of date).  For those using MM at least, creating a builder is very simple.  I agree that there is a lot of boiler plate code (but there's a lot of goodies in there :)).  If we can work to complete a generator, I think that will reduce concerns folks would have on the amount of code required to use the builder.  In the meantime, the MM templates should hopefully help.

jondavis9898:
Per discussion in #77, a new repo was created for this.  Submitted https://github.com/financialforcedev/fflib-apex-common-builder/pull/2 to the new repo.

This issue can be closed.

PR #103, opened 2016-02-06
Fix for System.TypeException thrown on orgs that have Spring '16. Cou…

rfjolnisson:
Fix for System.TypeException thrown on orgs that have Spring '16. Could be possible bug in release.

"System.TypeException: Collection store exception adding ktdev0.fflib_QueryFactory.InvalidFieldException to List<java.lang.Exception>
Stack trace: Class.ktdev0.fflib_QueryFactoryTest.invalidFieldTests: line 115, column 1"

### DIFF
--- a/fflib/src/classes/fflib_QueryFactoryTest.cls
+++ b/fflib/src/classes/fflib_QueryFactoryTest.cls
@@ -108,7 +108,7 @@ private class fflib_QueryFactoryTest {
 
 	@isTest
 	static void invalidFieldTests(){
-		List<Exception> exceptions = new List<Exception>();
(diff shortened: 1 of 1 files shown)


frup42:
Thanks for this!
Certainly looks like a platform issue, probably introduced in Spring 16 as you say. Have you reported it to Salesforce? Fix looks solid enough to me.

capeterson:
@rfjolnisson can you comment on if you're reported this to salesforce or not? It seems like the type of thing we should fix.

afawcett:
It has been reported and a known issue has been raised.

Issue #105, opened 2016-02-09
Poor performance of QueryFactory (most notably QueryField)


daveespo:
Sorry, it's going to be long. Brace yourself.

This is just a different version of what we reported in #79

We have the following frustration:
- If we try to debug Apex code (printf-style debugging; not the new Apex Interactive Debugger) with a level of FINEST, the log file fills up with a bajillion of these and overflows the 2MB log limit:

```
13:57:03.185 (1185547282)|METHOD_ENTRY|[EXTERNAL]|01p36000001tRgR|fflib_QueryFactory.QueryField.equals(Object)
13:57:03.185 (1185753260)|METHOD_EXIT|[EXTERNAL]|01p36000001tRgR|fflib_QueryFactory.QueryField.equals(Object)
```
- If we set a Trace Filter for fflib_QueryFactory to try to silence it (by setting Apex debug level to ERROR, not NONE .. due to the Known Issue mentioned in #79), we see these in the log instead:

```
13:59:44.970 (18970258739)|PUSH_TRACE_FLAGS|[EXTERNAL]|01p36000001tRgR|QueryField|APEX_CODE,ERROR;APEX_PROFILING,INFO;CALLOUT,INFO;DB,INFO;SYSTEM,DEBUG;VALIDATION,INFO;VISUALFORCE,INFO;WORKFLOW,INFO
13:59:44.970 (18970382445)|POP_TRACE_FLAGS|[EXTERNAL]|01p36000001tRgR|QueryField|APEX_CODE,ERROR;APEX_PROFILING,INFO;CALLOUT,INFO;DB,INFO;SYSTEM,DEBUG;VALIDATION,INFO;VISUALFORCE,INFO;WORKFLOW,INFO
```

Accompanying the log file overflowing, we often reach the Apex CPU Time Limit: "Apex CPU time limit exceeded. An unexpected error has occurred. Your development organization has been notified."

Our Selectors have some methods that retrieve _really_ wide records with lots of parent/grandparent fields included in the related field list .. sometimes a hundred fields wide.

We found a couple particularly painful hot spots in the code -- both are related to fflib_QueryFactory.QueryField.

[...]

PR #107, opened 2016-02-22
Removes need for QueryField custom Apex class per financialforcedev/f…

daveespo:
…flib-apex-common#105
#105


afawcett:
@daveespo  I think we can close this PR now, yes?

afawcett:
@daveespo I think we can also close this one as well now, yes?

PR #120, opened 2016-05-09
Added the User SOBject to list of Standard objects without the Curren…

monkeyhandz:
As reported in this [issue](https://github.com/financialforcedev/fflib-apex-common/issues/119), the `User` SObject is missing from the list of Standard Objects that do not have a `CurrencyISOCode` field.

### DIFF
--- a/fflib/src/classes/fflib_SObjectSelector.cls
+++ b/fflib/src/classes/fflib_SObjectSelector.cls
@@ -33,13 +33,7 @@ public abstract with sharing class fflib_SObjectSelector
     /**
      * This overrides the Multi Currency handling, preventing it from injecting the CurrencyIsoCode fie ld for certain System objects that don't ever support it
      **/
-    private static Set<String> STANDARD_WITHOUT_CURRENCYISO = new Set<String> { AsyncApexJob.SObjectType.getDescribe().getName()
-                                                                                , ApexClass.SObjectType.getDescribe().getName()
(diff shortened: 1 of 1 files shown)


fflib/src/classes/fflib_SObjectSelector.cls
```diff
@@ -33,7 +33,13 @@ public abstract with sharing class fflib_SObjectSelector
     /**
      * This overrides the Multi Currency handling, preventing it from injecting the CurrencyIsoCode fie ld for certain System objects that don't ever support it
      **/
-     private static Set<String> STANDARD_WITHOUT_CURRENCYISO = new Set<String> {'ApexClass', 'ApexTrigger', 'AsyncApexJob', 'Attachment', 'RecordType'};
+    private static Set<String> STANDARD_WITHOUT_CURRENCYISO = new Set<String> { AsyncApexJob.SObjectType.getDescribe().getName()
```
daveespo: This code change reverts https://github.com/financialforcedev/fflib-apex-common/issues/93 -- which was added to deal with a bizarre 'Entity is not api accessible' in PostInstall user context

Can you change these back to strings?

monkeyhandz: @daveespo Sorry I missed that the file had been changed. I've put in the change you requested.

afawcett:
Thanks! 👍

PR #128, opened 2016-07-15
PR for addQueryFactorySubselect depricated method

daveerickson:
https://github.com/financialforcedev/fflib-apex-common/issues/92

This was discussed and closed out earlier, but I don't see any evidence of the new methods in the project. The code coverage is a bit low on the fflib_SObjectSelectorTest but I'll see if I can improve that and get a PR in for that too.


frup42:
Hey, thanks for this contribution!

My only suggestion would be to add comments above the deprecated `addQueryFactorySubselect` methods pointing people to use your new versions, and a brief note on why it's a bad idea to use the potentially ambiguous SObjectType to resolve the relationship field on the parent query factory's SObject.

PR #132, opened 2016-07-21
Changed behavior on how currency ISO code fields are added to SOQL qu…

amarchbanks:
…eries. It is now done dynamically depending on whether the sObject has the field rather than creating a whitelist of sObjects which do not support the field.

### DIFF
--- a/fflib/src/classes/fflib_SObjectSelector.cls
+++ b/fflib/src/classes/fflib_SObjectSelector.cls
@@ -34,13 +34,14 @@ public abstract with sharing class fflib_SObjectSelector
      * Indicates whether the sObject has the currency ISO code field for organisations which have multi-currency
      * enabled. 
      **/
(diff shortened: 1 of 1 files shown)


afawcett:
Much  better thanks!

tmowbrey:
Hello,

I've updated the library since this was pulled into the repo and it appears as though CurrencyIsoCode is not being added automatically as expected. Is anyone else seeing this behviour as well?

afawcett:
@tmowbrey best raise this as a defect issue then, worth investigating for sure.

cropredy:
@amarchbanks Much appreciated as custom metadata types don't have `CurrencyIsoCode` and the whitelist would never have worked in that use case

afawcett:
@cropredy this PR has been merged, i think your saying we need a new issue for this issue?

cropredy:
yes - a new issue is required here; I was kind of hoping @AMarchbanks would have spotted this and raised/fixed it

PR #133, opened 2016-08-12
Update to API 37

daveerickson:
I needed access to Asset.RecordTypeId which wasn't accessible until API 33.


afawcett:
This looks like a duplicate PR, i'll close this one out and comment on the other. Thanks.

PR #134, opened 2016-08-12
Update to API 37

daveerickson:
I needed access to Asset.RecordTypeId which wasn't accessible until API 33.


afawcett:
So we typically update the entire library to all API versions, mixed problems can occur if its only partial. Can you submit a change for all API versions and run the tests please. Thanks.

daveerickson:
I haven't forgotten about this. I've requested a sandbox which has taken more than 2 days to refresh...
Once that's done I'll put the fflib-apex-mocks and fflib-apex-common classes in, upgrade the API's and test it out. I'll create pulls if all goes well.

afawcett:
No problem, this looks like your other PR is the one to merge, so i'll close this one.

PR #137, opened 2016-09-01
Fields that are polymorphic fail due to first type being used.

squattingdog:
When a polymorphic field is being used and the type chosen does not match the sObjectType set in the selector, the configureQueryFactoryFields method will fail.

Add the ability to set the sObjectType when validating the field exists on the object.  Use the type set in the selector to determine which polymorphic type to use.

Example: Lead.Owner.UserRoleID
Owner is GROUP | USER
Group is the first type returned and does not contain UserRoleID.  The method then fails when the getField result is null.

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/137)


frup42:
Actually, polymorphic fields in Salesforce are a special case.

**Lead.Owner** can relate to a **Queue** or a **User**.
But crucially, the queryable fields don't map to either Queue or User.

With a bit of experimenting and googling, it turns out that the related object type is an SObjectType called [Name](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/api/sforce_api_objects_name.htm). It's a coincidence that there are fields with the same name on the User and Name SObjectTypes. It would be a dangerous assumption to say all fields on the User SObjectType will also exist on Name.

So I think the fix is smaller, and shouldn't require any changes to method signatures. Just when trying to get the describe for the lookup referenceTo, it should be something like:
```SObjectType relatedSObjectType = relatedObjs.size() == 1 ? relatedObjs[0] : Name.SObjectType;```

**Comment by squattingdog** [comment]
In this case of Lead.Owner, the getDescribe method returns 2 values, Queue and User.  Name would not match the values returned.  The validation is performed using the defined selector's getSObjectType method to validate that the field exists on the specified sObjectType.  The UserSelector.getSObjectType returns User.getSObjectType if it were another selector, it would return the type specific to it.  

For the above reason, I am not following the "dangerous assumption" statement.  As stated, all the fields on User do not exist on Name.  If the relatedSObjectType is set to Name, the validation will still fail when looping the fields to find a User or Queue specific field.  For example, the field User.BadgeText does not exist on Name.  While this proposed solution is simpler on the surface, it still contains the same error this solution solves.  That is, when the process loops the fields for a given SObjectType, it may not find a valid match because the wrong type is being used for the matching process.

**Comment by frup42** [comment]
Hi again, I didn't entirely follow this:
> The validation is performed using the defined selector's getSObjectType method to validate that the field exists on the specified sObjectType. The UserSelector.getSObjectType returns User.getSObjectType if it were another selector, it would return the type specific to it.

But hopefully your own example will clarify things here.
> For example, the field User.BadgeText does not exist on Name

You are correct, User has a field called BadgeText, which Name does not.
That is why this SOQL works: `SELECT BadgeText FROM User`
But this SOQL fails: `SELECT Owner.BadgeText FROM Lead`
And this SOQL passes: `SELECT Owner.UserRoleId FROM Lead` - because of the Name.UserRoleId field, not because of the User.UserRoleId field.

While a specific Lead record's (i.e. a single DB row) Owner is either a USER or a GROUP, the Lead SObjectType's (i.e. the whole DB table) Owner lookup is related to the Name SObjectType, not  User or Group.

**Comment by frup42** [comment]
Ah, I follow a bit more now. You are using the UserSelector to ensure you select all interesting fields about the Lead's Owner. In fact, you need a NameSelector to configure the fields on the main QueryFactory instance.
```
//new Testfflib_UserSelector().configureQueryFactoryFields(qf, 'Lead.Owner');
new MyNameSelector().configureQueryFactoryFields(qf, 'Lead.Owner');
```

So I still feel the fix would be:
`SObjectType relatedSObjectType = relatedObjs.size() == 1 ? relatedObjs[0] : Name.SObjectType;`

**Comment by squattingdog** [comment]
Separate note here, the SF org user creds seem to be out of sync which is causing the validation deploy to fail. @afawcett

**Comment by squattingdog** [comment]
Hi David,
The issue that is being addressed is with the describe call.  It fails due to the incorrect SObjectType being used; the actual soql that is generated and ultimately executed is not in question as the error happens prior to any soql being generated.  If the field doesn't exist on Name then the describe still fails.

For example, if we are looping the values for Lead.Owner.BadgeText we get an array of [Owner, BadgeText] for the specific Lead sObjectType.  The first iteration through we get Owner.  On the describe we get back the array: [ Group, User].    Let's set the SObjectType to Name because there is more than 1 item in the array.  The next iteration we are describing Name to find BadgeText, however, BadgeText does not exist on Name.  This results in the same error as arbitrarily setting the sObjectType to the first item in the array.  

This PR addresses the issue by passing in the sObjectType from the concrete selector that is using the QueryFactory instance.  I understand that this creates a tighter coupling between the 2 classes.  Perhaps, passing an instance of the concrete SObjectSelector to the QueryFactory constructor and then subsequently calling the SObjectSelector instance's getSObjectType method is a bit better??

Interested to hear any proposals for this issue.

**Comment by afawcett** [comment]
@squattingdog i appreciate this is an old PR by now, wondered if you still had a few on my last question on the coupling? I think we are close to getting agreement and merging.

**Comment by squattingdog** [comment]
@afawcett hi, sorry for the delay, I lost track of this over the holidays.  Anyway, the current implementation of this PR passes the sObjectType.  Are you saying that is the best option at this time?

File: fflib/src/classes/fflib_QueryFactory.cls [hunk 1]
```diff
@@ -759,7 +788,15 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 			this.fieldName = fieldName;
 			this.setMessage( 'Invalid field \''+fieldName+'\' for object \''+objectType+'\'' );
 		}
-	}
+	}public class InvalidRelationshipException extends Exception{
```
agarciaffdc: @squattingdog sorry for asking this when your PR is opened since so long ago. But could you clean format a bit? This class should be at least one line below the `}` And we usually add open brace below the line instead of beside. For instance 
`}public class InvalidRelationshipException extends Exception{`
should be 
`}
public class InvalidRelationshipException extends Exception
{`

Don't worry about the code it is not yours. We will merge it as soon as you format it a bit.

agarciaffdc:
@dbtavernerffdc FYI

fflib/src/classes/fflib_QueryFactory.cls
squattingdog: yup, not sure how that happened...addressing those.

squattingdog:
FYI, there still seems to be an issue with the creds being used for the Travis CI builds.

/home/travis/build/financialforcedev/fflib-apex-common/lib/exec_anon.xml:147: INVALID_LOGIN: Invalid username, password, security token; or user locked out.

agarciaffdc:
Thanks for the hands up @squattingdog We will work around it in order to be able to merge it without Travis confirmation on your fork. Then Travis works properly on the lib repo

fflib/src/classes/fflib_QueryFactory.cls
```diff
@@ -760,6 +790,18 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 			this.setMessage( 'Invalid field \''+fieldName+'\' for object \''+objectType+'\'' );
 		}
 	}
+	
+	public class InvalidRelationshipException extends Exception
```
dbtavernerffdc: I can't find anywhere else in this pull request where this exception is referenced. Perhaps it should be thrown if the passed in object type is not in the reference to list.

fflib/src/classes/fflib_QueryFactory.cls
```diff
@@ -108,7 +108,22 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
...
 			if(token != null && i.hasNext() && tokenDescribe.getSOAPType() == Schema.SOAPType.ID){
-				lastSObjectType = tokenDescribe.getReferenceTo()[0]; //if it's polymorphic doesn't matter which one we get
+				if(relatedSObjectType == null) {
+					lastSObjectType = tokenDescribe.getReferenceTo()[0]; //if it's polymorphic get the first one - user did not specify the one to use.
+				}else{
+					List<Schema.sObjectType> relatedObjs = tokenDescribe.getReferenceTo(); //if it's polymorphic, it matters which one we use - i.e. Lead.Owner is GROUP|USER and each has different fields.
+					if(relatedObjs.size() > 1) {
+						String relatedSObjectName = fflib_SObjectDescribe.getDescribe(relatedSObjectType).getDescribe().getName();
+						for(Schema.sObjectType sot : relatedObjs) {
+							if(fflib_SObjectDescribe.getDescribe(sot).getDescribe().getName() == relatedSObjectName){
```
dbtavernerffdc: Out of interest why are we comparing the names and not just the sobject type objects?

fflib/src/classes/fflib_QueryFactory.cls
agarciaffdc: Good one!!

agarciaffdc: @squattingdog once you check this one we will merge it quickly.

JAertgeerts:
@afawcett @ImJohnMDaniel  can this be looked at? This is a pretty significant issue for any field (and increasing with the increase of polymorphic fields added with Industry clouds) that can have multiple objects as a related object.

JAertgeerts (approved):

PR #151, opened 2017-01-19
Fixes https://github.com/financialforcedev/fflib-apex-common/issues/150

up2go-rsoesemann:
The test addChildQueries_invalidChildRelationship() failed in orgs with relationship fields from Contact to Contact. This is case often seen. So I think using null instead of a specific type is less risky and tests exactly the same functionality.


frup42:
I agree with your logic, null seems like a safe bet to me!
Thanks!

PR #171, opened 2017-09-14
feat: Add optional parameter for registerDirty to specify exactly wha…

JonnyPower:
…t fields need to be updated

This allows a unit of work's uncommited changes to not be overwritten when multiple areas of code update the same record on the same unit of work.

e.g. a stack of processes that update different fields on the same record while being passed a unit of work allowing for changes to be commited after http callouts

Comes form Traction on Demand's John Rogers, has been sitting in our code base for a long time

There is a breaking change compared to a previous versions of fflib, where an excpetion is thrown if the same record is registered dirty twice without fields being specified. This could be modified to overwrite the existing registered dirty record (which is the current behaviour) instead of throwing an exception.

### DIFF
--- a/fflib/src/classes/fflib_ISObjectUnitOfWork.cls
+++ b/fflib/src/classes/fflib_ISObjectUnitOfWork.cls
@@ -60,6 +60,16 @@ public interface fflib_ISObjectUnitOfWork
      **/
     void registerDirty(SObject record);
     /**
+     * Register specific fields on record to be updated when work is commited
+     *
+     * If the record has previously been registered as dirty, the dirty fields on the record in this call will overwrite
+     * the values of the previously registered dirty record
+     *
+     * @param record An existing record
+     * @param dirtyFields The fields to update if record is already registered
+     **/
+    void registerDirty(SObject record, List<SObjectField> dirtyFields);
+    /**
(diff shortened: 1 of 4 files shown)


JonnyPower:
@afawcett any thoughts on this?

marktroupe:
This is a problem we've come across a lot. We've implemented our own work arounds in places, but nothing good enough to be permanent.

afawcett:
@JonnyPower sorry for the delay.

I like the concept of this registerDirty merge overload, nice work! As you say it does regress in terms of throwing an exception. As this is a core class we have to be sensitive to regressions, so if its ok, can i ask that you perform the following change? I'll be happy to approve then. 👍 

_"This could be modified to overwrite the existing registered dirty record (which is the current behaviour) instead of throwing an exception."_

JonnyPower:
@afawcett Sorry for taking forever to get back to this - there's a couple of issues I need to fix before this is ready to be merged. I'll reopen the PR when it's ready.

ryandraper:
Thanks for working on this @JonnyPower ! Can't wait til it's officially merged in 👍

mbuhler22:
Thanks for the hard work @JonnyPower ! We really need this feature.

Issue #199, opened 2018-09-14
Remove 'with sharing' from fflib_SObjectSelector and fflib_SObjectDomain


yurybond:
There is a big inconsistency in definition of selector base class and trailhead documentation of how it shoud be used. 
1. According to trailhead we need to avoid using of 'with/without sharing' on selectors to allow to apply calling context (for instance respect of controller's context when use selector method) See more:
https://trailhead.salesforce.com/en/modules/apex_patterns_dsl/units/apex_patterns_dsl_learn_selector_l_principles

2. For some base class for all selector classes  fflib_SObjectSelector  has 'with sharing' which is inherited and applied to child class (in other words selector that inherit base class fflib_SObjectSelector   will ignore caller 'sharing' context). See more
https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_classes_keywords_sharing.htm

3. I am afraid removing of 'with sharing' keyword from base selector class and base domain class will affect a lot of people around the world, but we have to do it to make it work properly.

jjulicher:
I totally agree with this. 
Although with the latest update perhaps this should be inherited sharing?

yurybond:
@jjulicher indeed! According to @afawcett  recommendation I created a new base class with inherited sharing. Andrew's comment in on my pull request: https://github.com/financialforcedev/fflib-apex-common/pull/200

wimvelzeboer:
I also agree that the base class should be set to `inherited sharing` so that the caller can decide how to use it. 

It would be nice to use the following pattern that I also use for Service classes.

**Application.cls**
```
public with sharing class Application
{
    public static final fflib_Application.SelectorFactory Selector =
            new fflib_Application.SelectorFactory(
                    new Map<SObjectType, Type>
                    {
                            Account.SObjectType => AccountsSelectorImp.class
                    }
            );

    public static final fflib_Application.SelectorFactory ElevatedSelector =
                new fflib_Application.SelectorFactory(
                        new Map<SObjectType, Type>
                        {
                                Account.SObjectType => AccountsSelectorElevatedImp.class
                        }
                );
}
```

**AccountsSelector.cls**
```
public abstract inherited sharing class AccountsSelector extends sflib_SObjectSelector
{
    ...
    public AccountsSelector()
    {
        super();
    }
    public AccountsSelector(Boolean includeFieldSetFields, Boolean enforceCRUD, Boolean enforceFLS)
    {
[...]

PR #200, opened 2018-09-14
#199 removed 'with sharing' keyword from base classes to respect call…

yurybond:
removed 'with sharing' keyword from base classes to respect caller sharing context

According to trailhead module about SOQ. The Selector should respect the context of the caller. It is not happening when base class has with/without sharing

> **Security** - Avoid using the with sharing or without sharing keywords on Selector Apex classes to ensure that the calling context inherits this context. Typically, this is the Service class, which, as per its design guidelines, must specify with sharing...  [see more](https://trailhead.salesforce.com/modules/apex_patterns_dsl/units/apex_patterns_dsl_learn_selector_l_principles)

@afawcett, @afawcettffdc


yurybond:
@dfruddffdc Could you please review this pull request?

yurybond:
@afawcett See your point about backward compatibility ) So have made changes in selector according your recommendations.

But, for the domain class, I even not sure how to deal with it. fflib_SObjectDomain has a lot of subtypes and static variables which make it almost impossible to move something to new Base class.

P.S. Is there somebody who can merge changes from pull requests?

fflib/src/classes/fflib_SObjectSelector2.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="urn:metadata.tooling.soap.sforce.com" fqn="fflib_SObjectSelector2">
+    <apiVersion>43.0</apiVersion>
```
jstorey-milcorp-ita:
```suggestion
    <apiVersion>44.0</apiVersion>
```

afawcett:
@yurybond any thoughts on the above?

yurybond:
@afawcett, 

1. I can roll-back version upgrade for package.xml and fflib_SObjectSelector2.cls.  As for fflib_SObjectSelectorBase.cls I am afraid we won't be able to save this class because `inherited sharing` was introduced in 43.

2. In terms of domain class: 
 - I guess it is better to make a changes in it in a separate branch
 - I spent a couple of hours trying to create 2nd version of domain with support of all constants and backward compatibility but it was pain - there are too many cross-references with other classes.

Andy, what do you think about making a fork of all project where we can re-design domain which is now looks messy and maintaining of it is a real pain? 

Also I have an idea to remove query builder from library. Query Builder was not applicable in projects with complex calculations because of high consumption of CPU time. The only benefit I saw there is checking access level to objects and fields, but shortly it will be supported by platform 
https://releasenotes.docs.salesforce.com/en-us/spring19/release-notes/rn_apex_select_with_security_enforced.htm

daveespo:
@yurybond  -- can you say more about the CPU issues with QueryFactory? A merge was made in 2017 that resolved a bunch of the overhead with using QueryFactory. It was resolved in #105 with PR #107 and Chris Peterson notes that QueryFactory is used not just for string concatenation but for handling Field Sets where there may be overlapping lists of fields across multiple FieldSets (https://github.com/financialforcedev/fflib-apex-common/issues/105#issuecomment-183587668)

I agree that QueryFactory should be retrofitted to use the new WITH SECURITY_ENFORCED but that sounds like a different issue :-)

resample-admin:
Can anyone comment on the status of this pull request?  

I would like the 'inherited sharing' behavior.  Wondering if I should modify my local files or if a a new version of fflib with them is coming soon.

daveespo:
@afawcett  -- I appreciate the breaking change that just changing SObjectSelector would cause, but with the changes to the Winter '20 Sites Guest User sharing model, many teams are now forced to more often run Apex in a Without Sharing context. I think we should either remove the sharing declaration (or use 'inherited sharing') for all classes.

A quick grep of the other Apex classes in this project and Apex Mocks shows that we have some 'with sharing' applied to classes that most certainly don't need it (fflib_StringBuilder, etc.) ... trying to exorcise those demons in a non-breaking-change way is much more complex. 

I think that we just need to explicitly call this out as a breaking change in the README -- we've made some behavioral changes back in 2014

CC: @agarciaffdc @dfruddffdc

daveespo:
One more comment on fflib_SObjectSelector: It's an abstract class ... it should really be up to the concrete implementation class to define the sharing model. Perhaps we should remove the sharing declaration altogether.

daveespo:
We're going to close this for now given the conversation over on #262

PR #209, opened 2018-11-16
Make UnitOfWork more extendable by adding virtual statements

wimvelzeboer:
By adding this change we make the UnitOfWork more flexible for extending it.

### DIFF
--- a/fflib/src/classes/fflib_SObjectUnitOfWork.cls
+++ b/fflib/src/classes/fflib_SObjectUnitOfWork.cls
@@ -141,7 +141,7 @@ public virtual class fflib_SObjectUnitOfWork
      * @param sObjectType - The type to register
      *
      */
-    private void handleRegisterType(Schema.SObjectType sObjectType)
+    private virtual void handleRegisterType(Schema.SObjectType sObjectType)
     {
         // add type to dml operation tracking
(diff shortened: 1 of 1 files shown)


afawcett:
Thanks for this @wimvelzeboer. There is already quite a bit of extensibility work in the class via **IDML** and **IDoWork** interfaces. Also the following methods... 

```    // default implementations for commitWork events
    public virtual void onRegisterType(Schema.SObjectType sObjectType) {}
    public virtual void onCommitWorkStarting() {}
    public virtual void onDMLStarting() {}
    public virtual void onDMLFinished() {}
    public virtual void onDoWorkStarting() {}
    public virtual void onDoWorkFinished() {}
    public virtual void onCommitWorkFinishing() {}
    public virtual void onCommitWorkFinished(Boolean wasSuccessful) {}```

The benifit of the above is they are expected extensibility points by the class. Its not to say that adding more would be an issue, but the core methods should i would say always have a predictable behavior. Whats your thoughts? Do any of the above methods help your use case?

afawcett:
@agarciaffdc i recommend we close the PR, we can always re-open it if @wimvelzeboer is keen to keep discussing. 👍

afawcett:
@agarciaffdc  this one can be closed in my opinion.

PR #216, opened 2019-01-29
Fix broken master branch

wimvelzeboer:
My change to the application didn't include the other merged change of the UnitOfWork. This will resolve that difference and make the fflib-apex-common repo deployable again.


agarciaffdc:
@wimvelzeboer Thanks for the hands up and the quick fix. Really appreciate.
cc @afawcett

PR #217, opened 2019-01-29
Some cleanup of UnitOfWork

wimvelzeboer:
I removed duplications of the validation checks, now the methods are more in one level of abstraction.
Also fixed some typos and some minor PDM stuff, mostly spaces and brace positions.

### DIFF
--- a/fflib/src/classes/fflib_SObjectUnitOfWork.cls
+++ b/fflib/src/classes/fflib_SObjectUnitOfWork.cls
@@ -40,7 +40,7 @@
  * "A Unit of Work keeps track of everything you do during a business transaction that can affect the database. When you're done,
  *  it figures out everything that needs to be done to alter the database as a result of your work."
  *
- * In an Apex context this pattern provides the following specific benifits
+ * In an Apex context this pattern provides the following specific benefits
  *  - Applies bulkfication to DML operations, insert, update and delete
  *  - Manages a business transaction around the work and ensures a rollback occurs (even when exceptions are later handled by the caller)
  *  - Honours dependency rules between records and updates dependent relationships automatically during the commit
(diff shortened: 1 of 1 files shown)


afawcett:
@agarciaffdc these looks good to me. I added a review as well.

fflib/src/classes/fflib_SObjectUnitOfWork.cls
```diff
@@ -628,14 +590,54 @@ public virtual class fflib_SObjectUnitOfWork
         }
     }
 
+	private void throwExceptionOnEventSObjectType(String sObjectType)
```
agarciaffdc: @wimvelzeboer nice enhancement. Is it possible to use another name on these methods? Something like checkException<blah> It confuses to me a bit. Name seems to say you will always throw an exception but this is not the case if the condition doesn't match.

wimvelzeboer:
@agarciaffdc & @afawcett I have updated this PR and also refactored the commitWork method, to improve readability.

fflib/src/classes/fflib_SObjectUnitOfWork.cls
```diff
@@ -120,13 +124,13 @@ public virtual class fflib_SObjectUnitOfWork
...
 
-        for(Schema.SObjectType sObjectType : m_sObjectTypes)
+        for (Schema.SObjectType sObjectType : m_sObjectTypes)
         {
             // register the type
             handleRegisterType(sObjectType);
         }
 
-        m_relationships.put( Messaging.SingleEmailMessage.class.getName(), new Relationships());
+		m_relationships.put(Messaging.SingleEmailMessage.class.getName(), new Relationships());
```
dbtavernerffdc: In the file view, this line looks too indented.

agarciaffdc:
@wimvelzeboer thanks for your latest contribution. Really appreciate it.

PR #225, opened 2019-05-17
Feature/avoid encrypted fields for default order in selector

wimvelzeboer:
This feature will prevent that the fflib_SObjectSelector class will use a Shield Encrypted Name field for the default ordering when no orderBy is provided.


wimvelzeboer:
Changes are already merged as part of another PR.

PR #227, opened 2019-05-21
Default Sorting considering invalid field and shield

tahsinz:
Financial Force library (FFLIB) by default sorts record based on CreatedDate. Some object (ie. account share) might not have the field and hence will through error.  This change checks the field encryption (taken from @wimvelzeboer ) and if the object does not have created date field, assigns Id as default sorting field.

Also, to ensure the inheritance for sobject selector (where FFLIB is used a different package or SFDX folder), certain variables has been changed from private to protected.

### DIFF
--- a/fflib/src/classes/fflib_SObjectSelector.cls
+++ b/fflib/src/classes/fflib_SObjectSelector.cls
@@ -69,7 +69,7 @@ public abstract with sharing class fflib_SObjectSelector
      * Switch this off if you need more performant queries.
      **/
     private Boolean m_sortSelectFields;
-  
+
    /**
    * Describe helper
    **/
@@ -81,7 +81,13 @@ public abstract with sharing class fflib_SObjectSelector
    		}
    		set;
    	}
-   	 
+   	/**
+   	* static variables
+    **/
+
+   	protected static String DEFAULT_SORT_FIELD = 'CreatedDate';
+   	protected static String SF_ID_FIELD = 'Id';
+
     /**
      * Implement this method to inform the base class of the SObject (custom or standard) to be queried
      **/
@@ -148,7 +154,7 @@ public abstract with sharing class fflib_SObjectSelector
 	 * Override this method to control the default ordering of records returned by the base queries,
 	 * defaults to the name field of the object if it is not encrypted or CreatedDate if there the object has createdDated or Id
 	 **/
-    public virtual String getOrderBy()
+     public virtual String getOrderBy()
     {
         if (m_orderBy == null)
         {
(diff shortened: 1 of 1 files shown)


tahsinz:
Great question and feedback @stohn777.  I have actually received an error regading this and hence added the logic to futureproof it :D I'm attaching the exception below:

I'm glad you liked the input regarding abstraction piece! I'm wondering what the next steps for merging the pull requests.

`Error on line 91, column 1: fflib_QueryFactory.InvalidFieldException: Invalid field 'CreatedDate' for object 'AccountShare'
Class.fflib_QueryFactory.getFieldPath: line 91, column 1
Class.fflib_QueryFactory.addOrdering: line 514, column 1
Class.fflib_SObjectSelector.configureQueryFactory: line 433, column 1
Class.GEN_SObjectSelector.newGenQueryFactory: line 82, column 1
Class.GEN_SObjectSelector.newGenQueryFactory: line 72, column 1
Class.CMT_AccountShareSelector: line 83, column 1`

stohn777:
Very well, then we need a bit of exception handling.  Thank you for the stack trace; that helped a lot.  Good job.

May I offer the following suggestion for a slightly cleaner look.  Just my minor thought.  :smiley:
```java
    public virtual String getOrderBy()
    {
        if (m_orderBy == null)
        {
            Schema.SObjectField nameField = describeWrapper.getNameField();
            if (nameField != null && !nameField.getDescribe().isEncrypted())
            {
                m_orderBy = nameField.getDescribe().getName();
            }
            else
            {
                m_orderBy = 'CreatedDate';
                try {
                    if (describeWrapper.getField(m_orderBy) == null)
                    {
                        m_orderBy = 'Id';
                    }
                }
                catch(fflib_QueryFactory.InvalidFieldException ex) {
                    m_orderBy = 'Id';
                }
            }
        }
        return m_orderBy;
    }
```

tahsinz:
Thanks, @stohn777 . I have cleaned up the code based on your suggestion and updated the pull request.

tahsinz:
@stohn777 I have updated the comments to fix the indenting issues (caused by IDE). Can you review it?

afawcett:
Thanks @tahsinz for this and @stohn777 for your sage advice here! :-) 

I think this looks good apart from the changes related to ```to ensure the inheritance for sobject selector (where FFLIB is used a different package or SFDX folder), certain variables has been changed from private to protected```, per discussion in the other PR, SFDX is not a use case for these changes, what use case is it specifically thats is driving the need for extensibility here? In the other PR it was adding some specific new features, what did you have in mind for the selector base class?

I would recommend meanwhile you remove those changes so we can get this PR merged and then discuss separately (can continue here if needed) the need for extensibility and how best to expose that based on your use case.

tahsinz:
Thanks @afawcett and I agree to your comments. I have updated the code as per your recommendation and I hope it is ready to be merged now.

dbtavernerffdc:
Code deploys to an org and tests all pass. Ready to merge.

PR #229, opened 2019-06-14
Upgrade API version for UnitOfWork to 46.0 to allow new object types

wimvelzeboer:
Object like ServiceResource and ServiceResourceSkill cannot be handled via UnitOfWork with API version 37.0, it results in the following error message: "Insufficient Privileges: You do not have the level of access necessary to perform the operation you requested."

Upgrading to API 46.0 resolves the issue


JAertgeerts:
@afawcett @pwrightffdc @stephenwillcock 
Are there any other individuals with write access that can do these peer reviews? Understand that the core team might not have the time, but the community should be able to drive some of these improvements forwards. 

Will open a main issue on this topic too.

afawcett:
Dupe with https://github.com/apex-enterprise-patterns/fflib-apex-common/pull/246

PR #232, opened 2019-06-27
Update fflib_SObjectDomain.cls

travis-rainey:
Do not return duplicate records if multiple fields are changed


afawcett:
Hmmm this might be a dupe with another PR i just noticed. Check.

travis-rainey:
> Hmmm this might be a dupe with another PR i just noticed. Check.

Yes sir...  Not sure how I missed that one but there is another PR for the same issue that was opened before mine.  I will close.

Issue #215, opened 2019-01-17
Update api version classes


dieffrei:
There some objets are not visible for fflib classes since the api version is out updated.
There are some standard objects like ServiceAppointment the QueryFactory doesn't work correctly and throw and error like: ServiceAppointment access is denied.
This error was really hard to figured out, then I think could be good update the api version of the classes

cropredyHelix:
Also fflib_queryFactory is unable to construct queries of PermissionSet for PermissionsXxxxx fields added since V37 like `PermissionsDeleteFieldHistory` which was added in V42.0 as file `fflib_SObjectDescribe.cls-meta.xml` is at v37.0

afawcett:
Done

PR #246, opened 2019-11-21
upgrade api version

JAertgeerts:
#215


foxcreation:
Nice modification! Is it an idea to move to API 47 directly? In addition, this will also allow the inParallel to test classes to speed up the deployment times.

afawcett:
I just updated it to 47 and ran all tests to confirm. @ImJohnMDaniel (or one other committer) can you add your review comments please and then we can get this merged.

Also... I think we should get as many outstanding PR's merged as possible in the current source format before moving to DX format. Otherwise it will just slow down getting these outstanding PR's merged.

PR #247, opened 2019-11-25
feature/mocking domain getChangedRecords

wimvelzeboer:
This change will allow us to use getChangedRecords in a unit-test that is using mocking.
It also now prevents duplicated records when more than one field is changed on the record.

As the Trigger.oldMap is readonly and directly referenced by this method you cannot use mocking for any method that has a reference to it.

Also did some re-formating and added two unit tests to cover the method.


afawcett:
@wimvelzeboer  your on a roll!

jhuizenga:
Nice one @wimvelzeboer! Was working this issue myself, but wasn't completely done yet. Came here to update my repo with the most recent changes, to find you allready done it 👍

Issue #233, opened 2019-07-30
List index out of bounds error on fflib_QueryFactory


wimvelzeboer:
Hi,
In some cases the fflib_QueryFactory will return a "List index out of bounds" error, thrown by the getFieldPath method when resolving a field on a related object e.g. `.selectField("Profile.Name")`, in combination with the users profile type and a combination of with and without sharing. It works perfect for Standard Salesforce License users.

This is a **unit test** which will replicate the issue:
```@IsTest
private with sharing class fflib_SObjectSelectorTest 
{
	@IsTest
	static void testElevatedSelector()
	{
		Id communityProfileId = [SELECT Id FROM Profile WHERE Name = 'Company Communities User' LIMIT 1].Id;
		User communityUser = new User(
				LastName = 'Smith',
				UserName = 'John.Smith@unittest.salesforce.com',
				Alias = 'JSmith',
				Email = 'John.Smith@unittest.salesforce.com',
				TimeZoneSidKey = 'America/Los_Angeles',
				LocaleSidKey = 'en_US',
				EmailEncodingKey = 'ISO-8859-1',
				LanguageLocaleKey = 'en_US',
				ProfileId = communityProfileId
		);
		insert communityUser;
		System.runAs(communityUser)
		{
			List<User> users = new UsersSelector().elevatedSelectByEmployeeNumber(new Set<String>{'123'});
		}
	}
}
```

And this is the **Selector** class:
```
public with sharing class UsersSelector extends fflib_SObjectSelector 
{
	public List<Schema.SObjectField> getSObjectFieldList()
	{
		return new List<Schema.SObjectField>
		{
				User.Id,
				User.Username,
				User.EmployeeNumber
		};
	}

	public Schema.SObjectType getSObjectType()
	{
		return User.SObjectType;
	}

	public List<User> elevatedSelectByEmployeeNumber(Set<String> employeenumbers)
	{
		ElevatedContext elevatedContext = new ElevatedContext(this);
		return  (List<User>) elevatedContext.selectByEmployeeNumber(employeenumbers);
	}

	private without sharing class ElevatedContext extends fflib_SObjectSelector
	{
		UsersSelector selector;

		public ElevatedContext(UsersSelector selector) {
			super(true, false, false, true);
			this.selector = selector;
		}

		public List<Schema.SObjectField> getSObjectFieldList() {
			return selector.getSObjectFieldList();
		}

		public Schema.SObjectType getSObjectType() {
[...]

PR #251, opened 2019-12-03
233 list out of bounds query factory

wimvelzeboer:
#233 Fix for list out of bounds error in QueryFactory when executed without sharing for a community user on an object that the user doesn't have access to.

When a query with a relationship fields in the select is generated for a community user in a without sharing context on an object that is not accessible (CRUD), a list out of bounds error is returned.

The method Schema.getGlobalDescribe().get('NonAccessibleObjectName__c').getDescribe().fields.getMap().get('MyRelationShipFieldToAnotherNonAccessibleObject__c').getDescribe().getReferenceTo() will return null when executed as community user.

While if we get the describe of that SObject for that relationship field via Schema.getGlobalDescribe().get('NonAccessibleRelationshipObjectName__c').getDescribe(), it will just return the expected describe.

So we have access to both SObjects, as expected because we run without sharing, but we do not have access to get the getReferenceTo on the relationship field between both SObjects.

As the purpose of this getFieldPath method is to validate the field level security for individual fields in the referred relationship, we can bypass this whole method when enforceFLS is disabled.

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/251)

### DIFF
--- a/fflib/src/classes/fflib_QueryFactory.cls
+++ b/fflib/src/classes/fflib_QueryFactory.cls
@@ -84,13 +84,19 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 	private Schema.ChildRelationship relationship;
 	private Map<Schema.ChildRelationship, fflib_QueryFactory> subselectQueryMap;
 
-	private String getFieldPath(String fieldName){
+	private String getFieldPath(String fieldName)
+	{
+		if (!enforceFLS)
+		{
+			// There is no need to validate FLS for individual fields in the relationship
+			return fieldName;
+		}
+
 		if(!fieldName.contains('.')){ //single field
 			Schema.SObjectField token = fflib_SObjectDescribe.getDescribe(table).getField(fieldName.toLowerCase());
 			if(token == null)
 				throw new InvalidFieldException(fieldName,this.table);
-			if (enforceFLS) 
-				fflib_SecurityUtils.checkFieldIsReadable(this.table, token);	
+			fflib_SecurityUtils.checkFieldIsReadable(this.table, token);
 			return token.getDescribe().getName();
 		}
 
@@ -103,7 +109,7 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 			Schema.SObjectField token = fflib_SObjectDescribe.getDescribe(lastSObjectType).getField(field.toLowerCase());
 			DescribeFieldResult tokenDescribe = token != null ? token.getDescribe() : null;
 			
-			if (token != null && enforceFLS) {
+			if (token != null) {
 				fflib_SecurityUtils.checkFieldIsReadable(lastSObjectType, token);
 			}
 


foxcreation:
@wimvelzeboer I'm not fully sure whether I agree with this change. In my perspective `getFieldPath()` performs more than just a FLS check. It also validates the field existence and, with that, ensures a distinct list of fields. By excluding that when FLS should not be asserted, one could end up having duplicate fields in the query (caused by typos).

```
fflib_QueryFactory qf = new fflib_QueryFactory(Contact.SObjectType);
qf.selectField('NAMe');
qf.selectFields( new Set<String>{'naMe', 'email'});
System.debug( '*** '+ qf.getSelectedFields() );
```
Result original code: `{Email, Name}`
Result modified code when FLS is false : `{NAMe, email, naMe}`

daveespo:
When I saw the code for this PR, I wanted to do the same test that @foxcreation did -- thanks for supplying the test case that illustrates why circumventing `getFieldPath()` has undesirable side effects

I agree that this PR can't be merged as-is 

The test case illustrated in #233  may be flawed: You can't change the sharing model of a class using inheritance. From the Apex docs: https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_classes_keywords_sharing.htm 

> Classes inherit this setting from a parent class when one class extends or implements another.

So I think the root cause of `getReferenceTo()` returning null is that `fflib_SObjectSelector` is running `with sharing` regardless of the `ElevatedSelector`'s sharing model

This is related to the discussion over on #199 -- I think we need to resolve that PR first before tackling this one 

@afawcett  -- keep me honest on the sharing model of subclasses 

cc @ImJohnMDaniel

daveespo:
I need to walk some of my previous comment back based on further experimentation we did.

In summary: The Apex docs are wrong. The sharing model used for executing SOQL is the sharing model defined on the class where the query actually runs. Inheritance doesn't come in to play at all. 

Consider the following base class example:

```
public virtual with sharing class BaseClass {
	public List<Contact> getContactsWithSharing(){
		return [SELECT Id, FirstName, LastName, Email FROM Contact];
	}
	
	public virtual List<Contact> getContactsUnclearSharing(){
		return [SELECT Id, FirstName, LastName, Email FROM Contact];
	}
}
```

and a class that derives from it:

```
public inherited sharing class OverrideClass extends BaseClass{
	public List<Contact> getContactsInhertedSharing(){
		return [SELECT Id, FirstName, LastName, Email FROM Contact];
	}
	
	public List<Contact> getContactsDelegateSharing(){
		return super.getContactsUnclearSharing();
	}
}
```

This is the result of calling methods from a WITHOUT SHARING context:

````
new OverrideClass().getContactsWithSharing(); //Returns zero rows (this is expected, according to the docs)
new OverrideClass().getContactsInheritedSharing(); //Returns all rows (see Note 1 below)
new OverrideClass().getContactsDelegateSharing(); //Returns zero rows (see Note 2 below)
````

Note 1: This is unexpected -- it directly contradicts the Apex Doc that says that "Classes inherit this setting from a parent class when one class extends or implements another." -- We're empirically observing that sharing model of the base class is irrelevant to the sharing model of methods defined in the subclass

Note 2: This is unexpected -- We're calling a method defined in a superclass and the sharing model is changing to the sharing model of the superclass

Based on the example that @wimvelzeboer provided on #233, he should not run afoul of a sharing-related restriction during construction of the QueryFactory so I have no explanation as to why the `getReferenceTo()` is coming back empty

@afawcett  and @capeterson  -- what's the process for getting a doc bug filed? And maybe an example to go with it :-)

wimvelzeboer:
@foxcreation Thank you for you insight in the usecases of the `getFieldPath()` method, and indeed I think we should make some changes to this PR.
@daveespo I think the issue is caused by something unforeseen by Salesforce, so I would think this is an actual bug in Salesforce.

@ImJohnMDaniel @afawcett would it be an idea to validate the output of `tokenDescribe.getReferenceTo()`. If it returns a null value and enforceFLS is disabled that we only in that scenario return the string 'fieldName'?

@afawcett Do you also agree that this is a bug in Salesforce?

capeterson:
@daveespo thanks for flagging. You can generally get a doc bug filed by just using the inline doc feedback form - that goes into a triage queue where the vast majority end up filed as bugs.

But since you flagged me down, no reason to trouble the doc writer triage queue, I'll make sure there's no exceptions to what you're showing and log the bug directly.

daveespo:
Super, thanks @capeterson

JAertgeerts:
@capeterson / @daveespo - has this bug been taken up or slated for future releases?

JAertgeerts:
This has forced us to apply the fix to the library in this commit. But would rather love salesforce to obviously resolve the internal.

cropredyHelix:
> @ImJohnMDaniel @afawcett would it be an idea to validate the output of tokenDescribe.getReferenceTo(). If it returns a null value and enforceFLS is disabled that we only in that scenario return the string 'fieldName'?

N.B. `tokenDescribe.getReferenceTo()` returns empty list, not null when running user doesn't have access to the object. There are other use cases besides the Community one; e.g. querying PermissionSetAssignment field `Assignee.Profile.Name` when user doesn't have `View All Data`

daveespo:
Per the comment on #309 and with Spring '21 and the #317 API upgrade, this should be resolved. Please reopen the issue if you find differently

PR #252, opened 2019-12-06
Semantic release management support

dieffrei:
I have been using fflib since 2013 it is great.
But one of commons complains about fflibs is about documentation and change-log.
This PR is about to automatic generate release notes and change log based on semantic commit messages.

More infos about it here:
https://github.com/semantic-release/semantic-release

The result after commit on master you can see:
https://github.com/dieffrei/fflib-apex-common/releases

It is necessary to set a specific variable to able to commit automatically to CI, feel free to reach me out.

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/252)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 *.prefs
 
 fflib/salesforce.schema
+.idea/
+node_modules/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,15 @@
-language: java
+dist: trusty
+language: node_js
+node_js:
+  - 10
 jdk:
   - oraclejdk8
 sudo: false
 branches:
   only:
     - master
-script: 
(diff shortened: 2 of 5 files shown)


afawcett:
@dieffrei what a great idea! Thank you. Will take a look.

dieffrei:
Thank you guys! Soon as possible I will submit a PR to help to lint commit
Messages in the right way

daveespo:
So much has changed since this PR has opened (we switched from Travis to GH Actions so this PR would require a full do-over) and we have been rather cautious over the years to not introduce breaking changes so the need to adhere to a semantic release versioning scheme is not something that's a priority. Thank you for your efforts but we will not be merging this PR.

PR #255, opened 2019-12-09
fix #254 issue

### LINKED ISSUE #254: fflib_SObjectDomain.getChangedRecords line 289 and 309, the changed record was added multiple times when this record has multiple fields changed
When a record has multiple fields changed, the two getChangedRecords method add it multiple times to changedRecords list.
I think it would be better to add a break statement after adding for line 289 and 309.

garychenming:
fix changed records are added multiple times issue in sobjectdomain


wimvelzeboer:
@garychenming thanks for your contribution! But I think we can decline this PR as this functionality is already included in the master branch via another PR.

garychenming:
@wimvelzeboer Okay. I'm glad to hear that this bug has been fixed in the mater branch.

daveespo:
Sounds like this is already resolved by another PR

PR #257, opened 2019-12-16
Feature/virtual application

wimvelzeboer:
Add the option to extend the fflib_Application class.

This will allow an extension that combines the force-di repo with the fflib-apex-commons


ImJohnMDaniel:
@wimvelzeboer -- Have you made your changes to hook into the Force-DI project yet?  If so, are they somewhere where we can review?

I have a similar example in the [AT4DX project](https://github.com/ImJohnMDaniel/at4dx/blob/master/sfdx-source/core/main/classes/framework-application-factory/Application.cls) that I would like to compare to yours.  I will be updating that version soon based on the changes supplied here.

Issue #163, opened 2017-04-08
fflib_QueryFactory support for toLabel()?


dudunato:
Is there a way to use QueryFactory and get the translation for a picklist? Like:

```
SELECT Id, Name, toLabel(ServiceType__c)
FROM Time__c
```

dudunato:
Nobody?

dudunato:
Anyone?

afawcett:
Sorry @dudunato for the delayed response. I don't see a way of doing this currently no. @capeterson and proposals on the best API design to accomplish this, maybe overload the select method with a boolean param 'toLabel' ?

dudunato:
Thanks for the reply @afawcett!

Wouldn't it be better if we would have the option on a field level?
Something like `query.selectField('toLabel(FieldName__c)')`

Or even `query.selectField('FieldName__c', true)` and this boolean parameter would be a flag to add the field to a Map to be converted to toLabel() when the method toSOQL be called.

Idk, just ideas!

afawcett:
Yes, your second idea is the one i was suggesting actually. Though your first is more flexible, it just means the query factory cannot validate the field as its an expression being passed.

tomjansen13:
Is there already a solution for this? I'm trying to get the label of a Data Category in a subquery on Knowledge Articles.

dudunato:
HI @tomjansen13,

I don't think this solutions was implemented. What I ended up doing was to use replace() function after building up the whole query. Super ugly, but... ¯\_(ツ)_/¯

tomjansen13:
Hi @dudunato Thanks for the 'solution'. I guess this is better than nothing at all!

anirudhs2005:
I was actually working on a similar requirement and I believe you can specify an alias to the field. SObject.get('<alias>') gives you the correct answer.

Example
1. List<SObject> records = Database.query('select id, Rating, toLabel(Rating) myAccountRating from Account LIMIT 1' );
2. String translatedValue = records[0].get('myAccountRating');
3. system.debug(translatedValue);

I am guessing it must be possible to add an alias to a toLabel field somewhere.

foxcreation:
Suggested PR change to allow toLabel within the fflib framework can be found here: https://github.com/apex-enterprise-patterns/fflib-apex-common/pull/258

PR #258, opened 2019-12-26
Enabling SOQL functions (#163)

foxcreation:
Enrichment to the fflib Query Factory to allow soql methods (like COUNT(), toLabel, and such) to be specified. Note, they can be added via `selectFields( String(s) )` methods, as is illustrated in the provided test methods.

### DIFF
--- a/fflib/src/classes/fflib_QueryFactory.cls
+++ b/fflib/src/classes/fflib_QueryFactory.cls
@@ -84,17 +84,51 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 	private Schema.ChildRelationship relationship;
 	private Map<Schema.ChildRelationship, fflib_QueryFactory> subselectQueryMap;
 
-	private String getFieldPath(String fieldName){
-		if(!fieldName.contains('.')){ //single field
-			Schema.SObjectField token = fflib_SObjectDescribe.getDescribe(table).getField(fieldName.toLowerCase());
-			if(token == null)
-				throw new InvalidFieldException(fieldName,this.table);
-			if (enforceFLS) 
-				fflib_SecurityUtils.checkFieldIsReadable(this.table, token);	
-			return token.getDescribe().getName();
+	/**
+	 * Pattern variable to prevent the Pattern to be compiled for each field
+	 */
+	private Pattern soqlFunctionPattern { get{
+		if( soqlFunctionPattern == null ){
+			soqlFunctionPattern = Pattern.compile( '^([a-zA-Z_]*\\(\\s*)([a-zA-Z0-9_\\.]+)?(\\s*\\)[a-zA-Z ]*)$' );
+		}
+		return soqlFunctionPattern;
+	} set; }
+
+	private String getFieldPath( String fieldName ){
+		// Check whether field contains a SOQL function statement - if so, split from field
+		String beforeField = '';
+		String afterField = '';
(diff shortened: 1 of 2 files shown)


fflib/src/classes/fflib_QueryFactory.cls
```diff
@@ -84,17 +84,51 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
...
+	 * Pattern variable to prevent the Pattern to be compiled for each field
+	 */
+	private Pattern soqlFunctionPattern { get{
+		if( soqlFunctionPattern == null ){
+			soqlFunctionPattern = Pattern.compile( '^([a-zA-Z_]*\\(\\s*)([a-zA-Z0-9_\\.]+)?(\\s*\\)[a-zA-Z ]*)$' );
+		}
+		return soqlFunctionPattern;
+	} set; }
+
+	private String getFieldPath( String fieldName ){
```
wimvelzeboer: By adding this logic inside this method the complexity suddenly becomes very high. I would advise to split this logic into a number of private methods to keep a level of abstraction here in his method.

foxcreation: @wimvelzeboer I do agree with you on the increased complexity and length of the method. I've thought of this before, but in my perspective this will only become more difficult to read, since beforeField and afterField is used throughout the function. The methods should then return a list or an innerclass to support. Global is also not an option since there will be multiple field paths called. That's why I've put emphasize on the comments.

If you have any suggestion how to split and make sure all data is properly fed back to be used, please let me know!

wimvelzeboer: I would maybe split up this method into three parts:
`private String getFieldPathForMethod(String fieldName)`
`private String getFieldPathForRelatedField(String fieldName)`
`private String getFieldPathForField(String fieldName)`

```
private String getFieldPath( String fieldName )
{
  if (fieldName.contains( '(' )) return getFieldPathForMethod(fieldName);
  if (fieldName.contains( '.' ) ) return getFieldPathForRelatedField(fieldName);
  return getFieldPathForField(fieldName);
}

private getFieldPathForMethod(String methodCall)
{
  Matcher match = this.soqlFunctionPattern.matcher( methodCall );
  if (!match.matches())
  {
    throw new InvalidFieldException( 'The method '+ methodCall + 'does specify a ( function opening, but doesn\'t match the pattern' );
  }
  
  String beforeField = match.group( 1 );
  String fieldName = match.group( 2 );
  String afterField = match.group( 3 );

  // When no fieldName is specified (e.g. COUNT() ), return directly, since nothing to check
  if( String.isBlank( fieldName ) )
  {
    return beforeField + afterField;
  }
  return beforeField + getFieldPath(fieldName) + afterField;
}
```

daveespo:
QueryFactory is not well suited for use in generating Aggregate SOQL queries. Sure, it can be bent to fit, but the primary driver behind the creation of QueryFactory was to enable Selector creation .. and those are tied 1:1 to SObjects. 

While direct use of QueryFactory is soemthing we know people do, we're not going to extend its use beyond the SObject primary use case. 

Additionally, with the introduction of native support for User Mode and System Mode in #419 , it is generally discouraged to use the old methods for unpacking the list of fields in the QueryFactory to individually test their readability via Describe methods. It's computationally expensive and something that UserMode handles natively. 

Thank you for your submission but we will not be merging this PR.

PR #268, opened 2020-02-17
Update fflib_SObjectSelector.cls

Javwad:
I think we can remove getSObjectType2() method and replace it with getSObjectType();

### DIFF
--- a/fflib/src/classes/fflib_SObjectSelector.cls
+++ b/fflib/src/classes/fflib_SObjectSelector.cls
@@ -330,15 +330,6 @@ public abstract with sharing class fflib_SObjectSelector
      * Public acccess for the getSObjectType during Mock registration 
      *   (adding public to the existing method broken base class API backwards compatability)
      **/
-    public SObjectType getSObjectType2()
-    {
-        return getSObjectType();
-    }
-
-    /**
-     * Public acccess for the getSObjectType during Mock registration 
-     *   (adding public to the existing method broken base class API backwards compatability)
-     **/
     public SObjectType sObjectType()
     {
         return getSObjectType();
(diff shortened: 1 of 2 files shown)


ImJohnMDaniel:
The fflib_SObjectSelectorTest.cls class would also need to be updated for this PR to be approved.

Javwad:
> The fflib_SObjectSelectorTest.cls class would also need to be updated for this PR to be approved.
@ImJohnMDaniel 
Updated fflib_SObjectSelectorTest.cls - replaced getSObjectType2() with getSObjectType(). I have checked in test class that we are using a child selector class which has extended the fflib_SObjectSelector class. Hence, test class will not throw "Method is not visible. - error"

Javwad:
> I don't understand the original intent of the getSObjectType2() method. I reviewed the work that originally created the method ([ec45ada](https://github.com/apex-enterprise-patterns/fflib-apex-common/commit/ec45ada13232acd69277daa8c8f3dacd2d1e29df)) and I'm wondering if it has something to do with masking .. i.e. there's an `sObjectType()` method and a `getSObjectType()` abstract method -- perhaps `getSObjectType()` can't be mocked and the `getSObjectType2()` is the way around that. This code all predates the introduction of the native Apex mocking framework so perhaps it's no longer necessary.
> 
> In short, I don't feel confident approving this PR.
> 
> @Javwad -- what problem are you trying to solve with this proposed change?

@daveespo I have the same question on the original intent of the method. I guess it's because we cannot call an abstract method from an abstract class. However, I am not facing any problem, it just seems like an additional step or piece of code which I believe we can remove.

ImJohnMDaniel:
@afawcett and @ca_peterson -- do either of you know the history of this `getSObjectType2()` method?

afawcett:
@Javwad @daveespo good detective work! It took me a few mins but yes, i do believe it was a way to access the selectors sobjecttype in a generic way when working with mocking. Per the comment on it it was done because elevating the abstract method to public appeared to have broken at the time the backwards compatibility of the class. Regardless though, there does appear to be a number of folks using this method (at least having done a global GitHub search). So if we are to maintain public API backwards compat even if its a bit less than ideal, we should probably just ignore this itch. https://github.com/search?q=getSObjectType2&type=Code

daveespo:
Good thinking with the Github search @afawcett  -- definitely not worth rocking the boat here

PR #272, opened 2020-03-05
Virtual SimpleDML to allow extending

wimvelzeboer:
Whenever the IDML interface gets updated, custom implementations of SimpleDML (instead of IDML) will be less impacted.

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls
@@ -93,25 +93,25 @@ public virtual class fflib_SObjectUnitOfWork
 	    void emptyRecycleBin(List<SObject> objList);
     }
 
-    public class SimpleDML implements IDML
+    public virtual class SimpleDML implements IDML
     {
-        public void dmlInsert(List<SObject> objList)
+        public virtual void dmlInsert(List<SObject> objList)
         {
             insert objList;
         }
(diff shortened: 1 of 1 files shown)


ImJohnMDaniel:
@wimvelzeboer -- What is the value add of making the SimpleDML methods virtual over just creating your own implementation of IDML??  You will be overriding the SimpleDML methods anyway, if you extend

wimvelzeboer:
Hi @ImJohnMDaniel,
We recently added a few methods to the IDML interface. That's when I found out the value of having an extendable SimpleDML. 
Now everyone that wanted something more than just the `SimpleDML implementation needed to create its own implementation from the `IDML` interface. 
If the `SimpleDML` was already extendable and when they had created their own implementation from `SimpleDML` then they didn't had to make any change to their existing source when that extra method was added to `IDML`. Then they just gained the added method.

Now we had add that `void emptyRecycleBin(List<SObject> objList)` method in a large number of the projects at my company. Which was basically a copy paste from what was on `SimpleDML`

wimvelzeboer:
@afawcett @stohn777 @daveespo @ImJohnMDaniel I resolved the merge conflict, can you review this. Would be very usefull to have this before the ExternalID feature is merged, as that also introduces a new method into the interface.

wimvelzeboer:
@afawcett Can you review this PR? We can really use this feature in my company

afawcett:
@wimvelzeboer done - thank you sir!

PR #273, opened 2020-03-06
Separate Domain from TriggerHandler

wimvelzeboer:
Ever since I started using fflib I felt like that the triggerHandler shouldn't belong inside the domain. With this change I have separated the domain class into a TriggerHandler class and an actual domain class.

To illustrate what kind of methods should go on the domain I added a few getter methods there.

@afawcett I saw in force-di that the trigger handler was also considered a separate thing, what are your thoughts on completely separating the handler from the domain? You will break backwards compatibility but gain a more elegant architecture.

The Diffs might show a bit odd, but I basically copy-passed all triggerHandler things from the domain class to a new class. Renamed a few variables/methods from *Domain* to *TriggerHandler*, no changes were made to its logic.

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/273)


stohn777:
@wimvelzeboer  Wow!  That's a lot of work, and high praise for your effort.  I, too, have been mulling the coupled nature of triggers and domains for quite some time although not having the opportunity to tackle it, so am very pleased to see your work.

I think this is going to be something that we digest and ruminate upon for a bit, for me at least, although I'm excited about the concept.

**Thinking out loud ...**
May we presume this is a breaking upgrade?  For those with extensive libraries of _old extensions_ it may be a daunting task to upgrade.  If we decide to adopt this separation, should we maintain the coupled version for a period of time to ease the transition while staying up to day?

wimvelzeboer:
@stohn777 Thanks for your feedback! Its was indeed quite some effort. And yes, this change (just added a few fixes) will break the backwards compatibility. I would like to start a discussion if we are OK with that to allow a better architecture.
If we would to give clear instructions on how to upgrade then I think we should be good.

To upgrade you need to make the following changes:

**Triggers**
Instead of referencing the domain they now need to change the class name to the trigger handler:
```
trigger ContactTrigger on Contact 
    (after delete, after insert, after undelete, after update, before delete, before insert, before update)
{
    fflib_SObjectTriggerHandler.triggerHandler(ContactTriggerHandler.class);
}
````
We might want to change the method name of `triggerHandler` as its now duplicating the class name. But didn't wanted to make that change yet to keep the old logic untouched.

***Domain to TriggerHandler**
Rename old domain classes like `Contacts.cls`  to `ContactsTriggerHandler.cls` and change the them to extend from `fflib_SObjectTriggerHandler` instead of `fflib_SObjectDomain`
Create a new domain class `Contacts.cls` extended from `fflib_SObjectDomain`

Now depending on how the domain class was used, they might need to move domain specific methods back to the newly created domain class. I actually expect that the domain was mostly used as trigger handler and then upgrading is just a matter of changing the class name and implementation.


So depending on the project and how fflib was implemented it can be quite easy or will require some effort and testing. 
You might even want to call this a new major version, like going from 1.0 to 2.0 :-P

**Comment by daveespo** [comment]
I'll admit that I didn't thoroughly review the breadth of this change -- but w/r/t to backwards compat: 

You'd never even be able to deploy this code into an org that already has Triggers that depend on `fflib_SObjectDomain.triggerHandler()` since you've removed that method signature entirely. That is a pretty large barrier to adoption.

Also, the set of steps for migration sound rather unreasonable.

I will agree that the coupling of Domain to the Trigger framework has always bothered me .. and for that reason, we don't use Domain classes much at all. Instead, we wrap most of our domain logic into Service classes. Our judicious use of Domain classes are _specifically_ when we're building something that is invoked by trigger and we want all of the conveniences that the current `fflib_SObjectDomain` provides. And most of the Domain methods delegate to Service classes quickly anyway. 

In summary, while I agree with the spirit of this suggestion, the collateral damage is too significant.

**Comment by wimvelzeboer** [comment]
@daveespo You have a point there with not being able to deploy the code. So I have added the backwards compatibility back in, by just keeping all the old methods and marking them as "deprecated".

In this way you can deploy the new change in an environment that is already using the triggerhandler on the domain. It will now send WARN messages to the user to migrate. Then at some point in the future we can remove the trigger handler from the domain.

What do you think of this?

**Comment by afawcett** [comment]
Yeah I do think that while the architecture is better we are bound to our signifiant customer base and avoiding disruption to that. Its great to support both API's, but this adds confusion and needs clarification in various blogs etc. I'm not convinced its really worth it to the end user, can we find some way to refactor internally and get the same benefits? 

I feel still bound to retaining the fact that Domain is both trigger logic and true business domain logic (per the common sample), even if those methods delegate to service layer in some cases. Its about encapsulation of all the objects behavior, not just its crud behavior. That said, you both seem to indicate this is something that as a pattern has ended up being less useful to you?

**Comment by wimvelzeboer** [comment]
>  That said, you both seem to indicate this is something that as a pattern has ended up being less useful to you?

@afawcett Indeed, the current architecture seems to be very confusing to developers. It is hard for them to see that the fflib_SObjectDomain is more than a trigger handler, that it is in the first place a Domain. Maybe this is also due to the lack of methods related to a domain, like the traditional getter, setter and filter methods.
I already added a few of those typical domain methods, maybe I should raise a separate pull-request with only those methods (and a few others). Then developers might start to use the Domain part of the fflib_SObjectDomain more often.

One idea to internally refactor would be to move the trigger logic methods into its own class, to separate it from the domain logic, and extend fflib_SObjectDomain from it, something like:
`public ...  class fflib_SObjectDomain extends fflib_SObjectTriggerHandler implements fflib_ISObjectDomain { ... }`. What are your thoughts on that?

**Comment by JAertgeerts** [comment]
I obviously echo Wim’s sentiment on this. The issue we found internally is that most developers coming from either a java background or no development background at all, have a lot of issues understanding what the domain does. For them, it’s like the good ol’ triggerhandler class and push all logic to a service method. 

From a versioning perspective, allowing breaking changes shouldn’t be a problem. Or we can spin up a new repo / tag where we tackle this architectural differences and apply them not caring about past use

**Comment by afawcett** [comment]
@wimvelzeboer i think this convo has flowed really well - we seem to be in place where backwards compatibility is not broken and with you latest idea the new base type gives a good SOC between domain and trigger handler code. @daveespo @ImJohnMDaniel @stohn777 thoughts?

**Comment by ImJohnMDaniel** [comment]
@wimvelzeboer -- your fork does not appear to be in sync with the main branch which is in DX Source Format now.  Can you incorporate the changes and let us know when you are ready and we can finish the review then?

**Comment by wimvelzeboer** [comment]
> @wimvelzeboer -- your fork does not appear to be in sync with the main branch which is in DX Source Format now. Can you incorporate the changes and let us know when you are ready and we can finish the review then?

@ImJohnMDaniel Fixed!

**Comment by wimvelzeboer** [comment]
@ImJohnMDaniel @afawcett @daveespo I have added a few extra accessor methods in fflib_SObjectDomain. Can you have a look at those? 
When this PR is merged, I also have a few additional things that I want to propose that are depended on this change.
Looking forward to your review

**Comment by daveespo** [comment]
@wimvelzeboer  -- I finally had a chance to go through this PR and pulled the branch locally to review it more detail. Additionally, I talked to @ImJohnMDaniel  and @afawcett IRT about the shape of this PR. 

The fact that the engineer is only subtly nudged to use the new `fflib_SObjectTriggerHandler` (via debug messages) is going to make the migration window effectively infinite. That is, existing code will run forever and as I'm sure we've all witnessed, most engineering teams don't have the management support for refactoring for the sake of refactoring. In short: we'll be forced to support BOTH `fflib_SObjectDomain` and `fflib_SObjectTriggerHandler` forever and the duplicated functionality it entails

Since maintaining backwards compatibility is so important for this library, we would like to propose something along the lines of having `fflib_SObjectDomain` extend your new `fflib_SObjectTriggerHandler`

ex:

```
public virtual with sharing class fflib_SObjectDomain extends fflib_SObjectTriggerHandler implements fflib_ISObjectDomain
```

That will allow you to move the trigger-related methods into `fflib_SObjectTriggerHandler` and leave the Domain-specific functionality in `fflib_SObjectDomain`

This seems like the best compromise between a pure design that separates concerns while maintaining backward compatibility

There are some other stylistic/best-practices things in this PR that I don't want to address just yet (ex. variable names should start with a lowercase letter; there are some risks to shadowing (`public Configuration Configuration` and `public static TestFactory Test`) .. I realize these are carry-overs from `fflib_SObjectDomain` but I'd like to take this opportunity to nudge people away from using those properties 

Lastly, @ImJohnMDaniel  and @afawcett  are in agreement that in parallel with these changes, revisions to existing documentation and new education materials should be a priority otherwise people will stick with the old documented patterns, unaware of what you're trying to accomplish here. 

cc @stohn777

daveespo:
@wimvelzeboer  -- checking in here -- we're trying to clean up unmerged PRs and if you're planning to refactor your proposed changes, I'll leave this one open. Otherwise, I'll close out this PR on Sept 7th (it can always get reopened)

wimvelzeboer:
Hi @daveespo,

I am looking into your comments. Its quite difficult to move the trigger handler methods to a separate class without breaking backwards compatibility. Many of the trigger handler methods are static in the fflib_SObjectDomain class, so they should than still stay or be forwarded.

Last day I have looked into an other interface structure; fflib_domain -> fflib_SObjects -> fflib_SObjectDomain (the current trigger handler), but still running into backwards compatibility issues.

Splitting the domain and triggerhandler is for us (Johnson & Johnson) very important, so any ideas are welcome. Maybe we can work on this together?

wimvelzeboer:
@ImJohnMDaniel @stohn777 @daveespo After our discussion yesterday I had a look again at this PR. Ideally I want to revert it back to a previous state, as this PR is basically duplicating code and only adding a deprecated message.
If we would redirect the deprecated methods to the new implementation then we could maintain backwards compatibility without duplicating source code.

daveespo:
@wimvelzeboer  -- you can either close this PR altogether and open a new one, or just push a completely fresh attempt at restructuring to this same branch and the PR will update. The comment thread on this PR is valuable but it may become irrelevant once you change the implementation.

daveespo:
Closing this PR since #300  is a more viable solution

PR #296, opened 2020-08-22
Upgrade the format for org preferences in the scratch org definition

PeterLinBiz:
Changes to be committed:
	modified:   config/project-scratch-def.json

https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_settings.htm

Re: https://github.com/orgs/apex-enterprise-patterns/projects/4
@stohn777

### DIFF
--- a/config/project-scratch-def.json
+++ b/config/project-scratch-def.json
@@ -4,9 +4,6 @@
     "settings": {
         "lightningExperienceSettings": {
             "enableS1DesktopEnabled": true
-        },
-        "mobileSettings": {
-            "enableS1EncryptedStoragePref2": true
         }
     }
 }
--- a/config/project-scratch-def.json
+++ b/config/project-scratch-def.json
@@ -2,8 +2,11 @@
     "orgName": "apex-common",
     "edition": "Developer",
     "settings": {
-        "orgPreferenceSettings": {
-            "s1DesktopEnabled": true
+        "lightningExperienceSettings": {
+            "enableS1DesktopEnabled": true
+        },
+        "mobileSettings": {
+            "enableS1EncryptedStoragePref2": true
         }
     }
 }


config/project-scratch-def.json
```diff
@@ -2,8 +2,11 @@
     "orgName": "apex-common",
     "edition": "Developer",
     "settings": {
-        "orgPreferenceSettings": {
-            "s1DesktopEnabled": true
+        "lightningExperienceSettings": {
+            "enableS1DesktopEnabled": true
+        },
+        "mobileSettings": {
+            "enableS1EncryptedStoragePref2": true
```
stohn777, line 9: Does an explicit need exist for the "mobileSettings" setting?

PeterLinBiz: Thanks for your question!
In my opinions, this is a not-so-easy-to-find but very useful (see the last link below **all three examples**) for Scratch Org Lightning Aura Components Development because it helps (when false) to disable caching setting during development as some links below. When it is disabled, we don't have to refresh the page several times just to see the new changes of page layouts or Lightning Aura Components during development.

S1EncryptedStoragePref2 in the OrgPreferenceSettings type (old format for API version less than 47.0) 
= enableS1EncryptedStoragePref2 (new format starting from API version 47.0)
= (Setup page at the first link below) the checkbox for “Enable secure and persistent browser caching to improve performance”

https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/debug_disable_caching.htm
https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_mobilesettings.htm
https://sfdctechie.wordpress.com/2019/12/24/turn-off-lightning-experience-caching-in-scratch-orgs/
https://releasenotes.docs.salesforce.com/en-us/winter20/release-notes/rn_api_meta_settings.htm
(It was included in **all three examples**.)  https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_settings.htm

stohn777: Good day @PeterLinBiz

Perhaps I'm missing something so please help me to understand.  😇

The linked documentation indicates the default value of the setting is "true", therefore this setting wouldn't be necessary nor change the shape of the scratch org.  If the intent is to disable the setting, then the specified value should be "false", overriding the default.

1. What is the intended outcome of the proposed change?  To explicitly enable the setting?  Or to disable the setting?
    1.1. If the intent is to explicitly enable the setting, what is the benefit for being explicit?
    1.2. If the intent is to disable the setting, then the proposed change is incorrect.  Although in that direction, why should a scratch org for AEP Common development disable the setting?

Thanks for clarifying this for me.  😄

PeterLinBiz: Good day @stohn777 !

* Regarding "enableS1DesktopEnabled":
I guess that "s1DesktopEnabled" (now new format "enableS1DesktopEnabled") was included in the original project-scratch-def.json probably because including it is convenient and straightforward as the reason below.
1. As link A and link B below, enableS1DesktopEnabled = Indicates whether Lightning Experience is turned on in the org (true) or not (false).
2. It seemed to me that we don't have to turn on Lightning Experience for AEP Common development.
3. It is okay to turn it on.

Link A: https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_lightningexperiencesettings.htm 
Link B: https://help.salesforce.com/articleView?id=lex_enable_turn_on.htm&type=5

* Regarding "enableS1EncryptedStoragePref2":
The reasons that I was proposing to add "enableS1EncryptedStoragePref2" as true are as below.
1. Yes, AEP Common development doesn't need this one to be false. It is true by default.
2. This fflib-apex-common is so useful for developers! That's why I so care of adding good thing to this library.
3. In my opinions, "enableS1EncryptedStoragePref2" is so useful for developers, but Salesforce didn't document it well enough as something "must-have" as the first two links in my previous comments above.
4. It is so convenient to set it as false on the fly to create a scratch org so that developers can gain the convenience of not refreshing pages several times. 
5. Because of 3 above, it is not-easy-to-find and so-nice-to-have.
6. There is no negative impact.
7. We may keep this good and convenient option with this excellent fflib-apex-common library and make it ready for someone really need it.
8. Probably we can add a few lines in README.md to explain why we want to add this.

@afawcett @ImJohnMDaniel @wimvelzeboer Would you like to consider adding this "enableS1EncryptedStoragePref2" ?

ImJohnMDaniel (approved):

PR #298, opened 2020-09-25
Allow for different implementations of Application factories

wimvelzeboer:
This PR is to create the possibility for different implementations of the Application factories.
It is fully backwards compatible with the static maps, as methods are replaced and directing to the new classes while showing a deprecated message with instruction on refactoring to the new classes.

This change is mainly done to allow for an implementation similar to force-di, or even make a direct link to it so that fflib can be used in combination with force-di.

The code is not yet entirely finished, but I mainly raised it now to open up the discussion to see the response on this idea.

### DIFF
--- a/docs/Application.md
+++ b/docs/Application.md
@@ -1,0 +1,45 @@
+FFLib Apex Common
+=================
+
+# Application class
+
+### Example - Binding based on static maps in Apex
+
+```apex
+public with sharing class Application
+{
+    public static final fflib_ServiceBindingResolver Service =
+            new fflib_ClassicServiceBindingResolver(
+                    new Map<Type, Type>
+                    {
+                            AccountsService.class => AccountsServiceImp.class
+                    }
+            );
+
+    public static final fflib_SelectorBindingResolver Selector =
+            new fflib_ClassicSelectorBindingResolver(
+                    new Map<SObjectType, Type>
+                    {
+                            Account.SObjectType => AccountsSelector.class
+                    }
+            );
+
+    public static final fflib_DomainBindingResolver Domain =
+            new fflib_ClassicDomainBindingResolver(
+                    Application.Selector,
+                    new Map<SObjectType, Type>
+                    {
+		                    Schema.Account.SObjectType => Accounts.Constructor.class
+                    }
+            );
+
+    public static final fflib_Application.UnitOfWorkFactory UnitOfWork =
+            new fflib_Application.UnitOfWorkFactory(
+                    new List<SObjectType>
+                    {
+                            Account.SObjectType,
+                            Contact.SObjectType
+                    }
+           );
+}
+```
--- a/sfdx-source/apex-common/application/classes/implementation/fflib_ClassicDomainBindingResolver.cls
+++ b/sfdx-source/apex-common/application/classes/implementation/fflib_ClassicDomainBindingResolver.cls
@@ -1,0 +1,129 @@
+/**
+ * File Name: fflib_ClassicDomainBindingResolver
+ * Description: Domain class binding resolver based on the classic AEP 1.0 definition with static maps
+ * Copyright (c) 2020 Johnson & Johnson
+ *
+ * @author : architect ir. Wilhelmus G.J. Velzeboer | wvelzebo@its.jnj.com
+ */
+public virtual class fflib_ClassicDomainBindingResolver implements fflib_DomainBindingResolver
+{
+	protected fflib_SelectorBindingResolver selectorBindingResolver;
+
+	protected Map<Schema.SObjectType, Type> domainConstructorBySObjectType;
+
+	protected Map<Schema.SObjectType, fflib_ISObjectDomain> mockDomainInstanceBySObjectType;
+
+	/**
+	 * Class constructor
+	 *
+	 * @param selectorBindingResolver
+	 * @param domainConstructorBySObjectType
+	 */
+	public fflib_ClassicDomainBindingResolver(
+			fflib_SelectorBindingResolver selectorBindingResolver,
+			Map<SObjectType, Type> domainConstructorBySObjectType)
+	{
+		this.selectorBindingResolver = selectorBindingResolver;
+		this.domainConstructorBySObjectType = domainConstructorBySObjectType;
+		this.mockDomainInstanceBySObjectType = new Map<SObjectType, fflib_ISObjectDomain>();
+	}
+
+	/**
+	 * Dynamically constructs an instance of a Domain class for the given record Ids
+	 *   Internally uses the Selector Factory to query the records before passing to a
+	 *   dynamically constructed instance of the application Apex Domain class
+	 *
+	 * @param recordIds A list of Id's of the same type
+	 * @exception Throws an exception via the Selector Factory if the Ids are not all of the same SObjectType
+	 *
+	 * @return Instance of fflib_ISObjectDomain containing the record with the provided Ids
+	 **/
+	public virtual fflib_ISObjectDomain newInstance(Set<Id> recordIds)
+	{
+		return newInstance(selectorBindingResolver.selectById(recordIds));
+	}
+
+	/**
+	 * Dynamically constructs an instance of the Domain class for the given records
+	 *   Will return a Mock implementation if one has been provided via setMock
+	 *
+	 * @param records A concrete list (e.g. List<Account> vs List<SObject>) of records
+	 * @exception Throws an exception if the SObjectType cannot be determined from the list
+	 *              or the constructor for Domain class was not registered for the SOBjectType
+	 *
+	 * @return Instance of fflib_ISObjectDomain containing the provided records
+	 **/
+	public virtual fflib_ISObjectDomain newInstance(List<SObject> records)
+	{
+		SObjectType domainSObjectType = records.getSObjectType();
+		if (domainSObjectType == null)
+			throw new DeveloperException('Unable to determine SObjectType');
+
+		// Mock implementation?
+		if (mockDomainInstanceBySObjectType.containsKey(domainSObjectType))
+			return mockDomainInstanceBySObjectType.get(domainSObjectType);
+
+		// Determine SObjectType and Apex classes for Domain class
+		Type domainConstructorClass = domainConstructorBySObjectType.get(domainSObjectType);
+		if (domainConstructorClass == null)
+			throw new DeveloperException('Domain constructor class not found for SObjectType ' + domainSObjectType);
+
+		// Construct Domain class passing in the queried records
+		fflib_SObjectDomain.IConstructable domainConstructor =
+				(fflib_SObjectDomain.IConstructable) domainConstructorClass.newInstance();
+		return (fflib_ISObjectDomain) domainConstructor.construct(records);
+	}
+
+	/**
+	 * Dynamically constructs an instance of the Domain class for the given records and SObjectType
+	 *   Will return a Mock implementation if one has been provided via setMock
+	 *
+	 * @param records A list records
+	 * @param domainSObjectType SObjectType for list of records
+	 * @exception Throws an exception if the SObjectType is not specified or if constructor for Domain class was not registered for the SObjectType
+	 *
+	 * @remark Will support List<SObject> but all records in the list will be assumed to be of
+	 *         the type specified in sObjectType
+	 *
+	 * @return Instance of fflib_ISObjectDomain containing the provided records
+	 **/
+	public virtual fflib_ISObjectDomain newInstance(List<SObject> records, SObjectType domainSObjectType)
+	{
+		if (domainSObjectType == null)
+			throw new DeveloperException('Must specify sObjectType');
+
+		// Mock implementation?
+		if (mockDomainInstanceBySObjectType.containsKey(domainSObjectType))
+			return mockDomainInstanceBySObjectType.get(domainSObjectType);
+
+		// Determine SObjectType and Apex classes for Domain class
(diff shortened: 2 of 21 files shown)


ImJohnMDaniel:
@wimvelzeboer -- Can you compare how this approach is any different than the Application class/approach found in the AT4DX project?  That Application already uses the [Force-DI](https://github.com/apex-enterprise-patterns/force-di) project and does not introducing a dependency from [fflib-apex-common](https://github.com/apex-enterprise-patterns/fflib-apex-common) to [Force-DI](https://github.com/apex-enterprise-patterns/force-di)
* [AT4DX Application class](https://github.com/apex-enterprise-patterns/at4dx/blob/master/sfdx-source/core/main/classes/framework-application-factory/Application.cls)

wimvelzeboer:
@ImJohnMDaniel It is kinda similar in what is tries to do, except this design is more on a higher level. AT4DX tries to wrap around some of the interface structure limitations. 
For example; you have framework specific source-code in the Application class (e.g. Application.SelectorFactory). In my opinion the Application class should belong to the application and the selectorFactory is something of the framework. Combining the two different pieces of code is confusing for developer.
In this PR, the issue is solved by improving the interface structure if the framework itself, and allowing for different implementations on a higher level of abstraction.
The lack of that improved interface structure results into a lot of complexity as you can see in AT4DX.

I also prefer to have everything in one framework and not having to install too many sub-libraries. So, I think either AT4DX should be added to apex-common or something like this PR.

ImJohnMDaniel:
>  In my opinion the Application class should belong to the application and the selectorFactory is something of the framework.

I would agree with you.  In the traditional implementation of the Application class (non-DX/2GP concerns), it does have that distinct separation.  

Take a look at the [Application.cls](https://github.com/apex-enterprise-patterns/fflib-apex-common-samplecode/blob/master/sfdx-source/apex-common-samplecode/main/classes/Application.cls) found in the [fflib-apex-common-samplecode project](https://github.com/apex-enterprise-patterns/fflib-apex-common-samplecode).  In that version, the Application.cls is completely specific to the application you are building and relies on the framework pieces from the fflib_Application.cls (found in this repo).

The AT4DX approach does move the Application class under the "framework" umbrella but every mapping is now dynamic in the [ApplicationFactory CMDT objects](https://github.com/apex-enterprise-patterns/at4dx/tree/master/sfdx-source/core/main/schema/objects).  Each application (aka 2GP) inserts their own mappings into the CMDT object so that every application in the org can use the same, common convention of a "Application" class without the need to modify the mappings in the Application class directly (as is the situation in the classic implementation of the Application factory).

> I also prefer to have everything in one framework and not having to install too many sub-libraries. So, I think either AT4DX should be added to apex-common or something like this PR.

While I hear you on this point, you are advocating the merging of potentially four frameworks -- Apex Mocks, Apex Common, Force-DI, and AT4DX.  For those using Apex Common in the traditional way, they already have the sub-library of Apex Mocks and if we were to merge, there would be backward-compatibility issues around AT4DX's introduction of the Application.cls verses developers own, classic implementation of Application.cls

Granted, I need to take a deeper dive into the changes you are suggesting in this PR, so please give me some time to digest that.  I should have some time in the coming days to do this.

Also, can you elaborate further on why the AT4DX Application class implementation confuses developers?

wimvelzeboer:
> Also, can you elaborate further on why the AT4DX Application class implementation confuses developers?

I think its mainly the different levels of abstraction. AT4DX is basically introducing a complete new Application factory / feature, duplicating the one from apex-common. While the OO principles would be more to keep the single feature / interface but with different implementations. Similar as the design principle of Force-DI, one feature-interface with multiple implementations.

ImJohnMDaniel:
Ok, I see what you are saying.  It was the decision to not alter the fflib-apex-common UOW and other base classes to ensure backwards compatibility for the classic fflib-apex-common implementations.  

@afawcett, @daveespo, and @stohn777 -- I would like you three to also take a deep look at this PR and add your comments, please.

stohn777:
@wimvelzeboer 
I'm going to take a look at this, but for starters, could you please address the conflicts as you see fit respecting other changes to 'master'.

wimvelzeboer:
> @wimvelzeboer I'm going to take a look at this, but for starters, could you please address the conflicts as you see fit respecting other changes to 'master'.

Yes, it is indeed a bit outdated looking at the changes after I raised this PR. Let me work on rebasing and updating the code!

wimvelzeboer:
@stohn777 I saw that there was a lot of overlap with other changes that were already accepted. So, I did some refactoring of those. Now this PR mainly includes the following:

Adds more documentation to Interface classes.

Adding of replaceWith method to factories
With this you can replace implementations in real-time. Very useful when you are doing AB testing or dynamic feature toggling.

Implement SOLI**D** principles in the Application class, 
which only references interfaces and not real implementations to make it easy to replace any implementation.

Adds extra setMock overloads to make mocking easier.
Instead of doing this:
```
fflib_ApexMocks mocks = new fflib_ApexMocks();
AccountsSelector selectorMock = (AccountsSelector) mocks.mock(AccountsSelector.class);
mocks.startStubbing();
mocks.when(selectorMock.sObjectType()).thenReturn(Schema.Account.SObjectType);
mocks.stopStubbing();
Application.Selector.setMock(selectorMock);
```
You can now do the same without the need to stub it:
```
fflib_ApexMocks mocks = new fflib_ApexMocks();
AccountsSelector selectorMock = (AccountsSelector) mocks.mock(AccountsSelector.class);
Application.Selector.setMock(Schema.Account.SObjectType, selectorMock);
```

stohn777 (changes requested):
I'm not persuaded with the need for `Application.Selector.setMock(SObjectType, fflib_ISObjectSelector)`.  What is the benefit for adding a mock thing that hasn't been properly mocked, other than saving 3 lines of code which is important to the functionality of the Application's thing factories' internal maps?

> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/298#-:-N-ny9MGBayZ7ZxpgKnN:b1c2i4g)* status: 0 of 25 files reviewed, 1 unresolved discussion (waiting on @afawcett, @daveespo, @dbtavernerffdc, @ImJohnMDaniel, @kjonesffdc, @swillcockffdc, and @wimvelzeboer)

*[a discussion](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/298#-N-nv5sc4sSrRM3kAT0C:-N-nv5sc4sSrRM3kAT0D:b5wl5py) (no related file):*
Let me see if I got this right. This PR intends to offer a way to register and get a thing (Selector, Domain, etc.) from a SObjectType that is different from the SObjectType related to the Ids in the parameter.  For instance, one might provide a list of Ids for child SObjects to the parent's thing which might select the related parents.   Is my understanding up to par?

If so, I'd like the java-doc comments to explain that more explicitly.

If not, then please help me to understand what the intent of these methods.

The intent of the `replaceWith` methods is understandable, the others are less clear.

---

wimvelzeboer:
> I'm not persuaded with the need for `Application.Selector.setMock(SObjectType, fflib_ISObjectSelector)`. What is the benefit for adding a mock thing that hasn't been properly mocked, other than saving 3 lines of code which is important to the functionality of the Application's thing factories' internal maps?

I did find a lot of code duplication in many unit-test where the only thing that was being stubbed was the SObjectType. Maybe the use-case is mainly with Domains than Selectors, as Domains do not always have anything to return. You then basically only want to verify that methods are invoked.

> [a discussion](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/298#-N-nv5sc4sSrRM3kAT0C:-N-nv5sc4sSrRM3kAT0D:b5wl5py) (no related file):
Let me see if I got this right. This PR intends to offer a way to register and get a thing (Selector, Domain, etc.) from a SObjectType that is different from the SObjectType related to the Ids in the parameter. For instance, one might provide a list of Ids for child SObjects to the parent's thing which might select the related parents. Is my understanding up to par?

The intend of the `List<SObject> selectById(Set<Id> recordIds, SObjectType sObjectType)` method, is to provide the SObjectType of the Ids. So that fflib doesn't need to determine the SObjectType and validate the Ids again.
Like in the context of a domain class:
```
public class Accounts extends fflib_SObjects implements IAccounts {
  public static IAccount newInstance(Set<Id> ids) {
    return (IAccounts) Application.newInstance(ids, Schema.Account.SObjectType);
  }
}
```
The main reason for this was to avoids run-time errors, in the case where the Id set is empty, when you do constructs like:
```
IContacts contactsDomain = Contacts.newInstance(contactRecordsWithoutAccount);
IAccounts domain = Accounts.newInstance(
    contactsDomain.getAccountIds()
);
```
Adding `isEmpty()` checks all over the place might become cumbersome, therefore we came up with this method providing that SObjectType to avoid the run-time exceptions.

stohn777 (changes requested):
Continuing in line.

> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/298#-:-N0MjI413irs8AduafCe:bc8o9ti)* status: 0 of 25 files reviewed, 4 unresolved discussions (waiting on @afawcett, @daveespo, @dbtavernerffdc, @ImJohnMDaniel, @kjonesffdc, @swillcockffdc, and @wimvelzeboer)

*[a discussion](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/298#-N-nv5sc4sSrRM3kAT0C:-N0Mj5q_71fs95KwghG-:bc8o9ti) (no related file):*

Let me see if I got this right. This PR intends to offer a way to register and get a thing (Selector, Domain, etc.) from a SObjectType that is different from the SObjectType related to the Ids in the parameter. For instance, one might provide a list of Ids for child SObjects to the parent's thing which might select the related parents. Is my understanding up to par?

The intend of the `List<SObject> selectById(Set<Id> recordIds, SObjectType sObjectType)` method, is to provide the SObjectType of the Ids. So that fflib doesn't need to determine the SObjectType and validate the Ids again.
Like in the context of a domain class:
```
public class Accounts extends fflib_SObjects implements IAccounts {
  public static IAccount newInstance(Set<Id> ids) {
    return (IAccounts) Application.newInstance(ids, Schema.Account.SObjectType);
  }
}
```
The main reason for this was to avoids run-time errors, in the case where the Id set is empty, when you do constructs like:
```
IContacts contactsDomain = Contacts.newInstance(contactRecordsWithoutAccount);
IAccounts domain = Accounts.newInstance(
    contactsDomain.getAccountIds()
);
```
Adding `isEmpty()` checks all over the place might become cumbersome, therefore we came up with this method providing that SObjectType to avoid the run-time exceptions.


 

</blockquote></details>

Continuing in line.

---
___
*[sfdx-source/apex-common/main/classes/fflib_Application.cls, line 283 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/298#-N0MnE-g0uU6GIphyh_2:-N0MnE-g0uU6GIphyh_3:b-j6o7qn) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/f4ced88c7e961aa5ac01d2629a04e2d3cad07b51/sfdx-source/apex-common/main/classes/fflib_Application.cls#L283)):*
> ```VBA
> 		 */
> 		public virtual List<SObject> selectById(Set<Id> recordIds, SObjectType sObjectType)
> 		{
> ```

This if statement seems to block the intended use case, instantiating a Selector with a blank set.
___
*[sfdx-source/apex-common/main/classes/fflib_Application.cls, line 339 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/298#-N0Mix6XBasREmgToPyj:-N0Mix6XBasREmgToPyk:b-a824s2) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/f4ced88c7e961aa5ac01d2629a04e2d3cad07b51/sfdx-source/apex-common/main/classes/fflib_Application.cls#L339)):*
> ```VBA
> 		 *                    avoids the need to stub the mock to return its SObjectType
> 		 * @param selectorInstance The instance of the mocked selector
> 		 */
> ```

With "fflib_Application" being virtual, I believe a convenience method may exist in the concrete implementation, if needed.
___
*[sfdx-source/apex-common/main/classes/fflib_Application.cls, line 420 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/298#-N0MjNfn5LqCbdLb__T8:-N0MjNfn5LqCbdLb__T9:blvumwg) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/f4ced88c7e961aa5ac01d2629a04e2d3cad07b51/sfdx-source/apex-common/main/classes/fflib_Application.cls#L420)):*
> ```VBA
> 		 * @return Instance of a Domain containing the queried records
> 		 * @exception Throws an exception via the Selector Factory if the Ids are not all of the same SObjectType
> 		 **/
> ```

In our conversation above, the use case is defined as "calling this newInstance method with an empty set and SObjectType.  The benefit is avoiding a runtime exception.  However as I follow from this entry point, a runtime exception would still be generated on line 285 because the set is empty.  The code on line 287, using the SObjectType, would not be reached.
```
FATAL_ERROR|fflib_Application.DeveloperException: Invalid record Id's set

Class.fflib_Application.SelectorFactory: line 285, column 1
Class.fflib_Application.DomainFactory: line 423, column 1
AnonymousBlock: line 1, column 1

Executed with:
    AccountsDomain accounts = (AccountsDomain) Application.Domains.newInstance(new Set<Id>(), Account.SObjectType);
```
Seemingly the if statement is unnecessary in that context, at line 284.

PR #300, opened 2020-10-12
Add new Domain structure

wimvelzeboer:
Domain structure based on Objects instead of SObject to allow for DTO, compound domains and other fun stuff.

@daveespo @ImJohnMDaniel @stohn777 After our discussion I gave it some more thought. And maybe we should flip it around. That we keep the current combined domain plus triggerHandler more-or-less as-is, but that we add a new domain structure above fflib_SObjectDomain and extended it from the new domain.

In this manner you have (almost) 100% backwards compatibility, except for the popTriggerInstance method. I am open to any thoughts on how to best solve that.

Eventually you can deprecate the entire fflib_SObjectDomain class when we have this new Domain structure combined with the future rebuild of the triggerHandler. 
I thought that something like this would keep the source much cleaner without having to redirect everything from old to new and add tons of deprecated messages.

Something that still need to be done in this PR is to modify the fflib_Application to accept fflib_Domain instead of fflib_SObjectDomain, but wanted to share this first to continue the discussion.

The new Interface / implementation structure would then look like this:
- fflib_Domain       (Main interface to define a domain)
- fflib_Objects       (Top level Domain class implementation with generic Objects)
- fflib_SObjects     
Similar to the current fflib_SObjectDomain with standard methods typical for domains. No trigger handler here.
New domain implementation for standard or custom objects should then be extended from this instead of fflib_SObjectDomain e.g. `public with sharing class AccountsImp extends fflib_SObjects implements Accounts`
- fflib_SObjectDomain        for backwards compatibility

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_Objects.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_Objects.cls
@@ -1,0 +1,108 @@
+/**
+ * Copyright (c), FinancialForce.com, inc
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ *   are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *      this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *      this list of conditions and the following disclaimer in the documentation
+ *      and/or other materials provided with the distribution.
+ * - Neither the name of the FinancialForce.com, inc nor the names of its contributors
+ *      may be used to endorse or promote products derived from this software without
+ *      specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ *  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ *  OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ *  THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ *  EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ *  OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ *  OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+**/
+public virtual class fflib_Objects implements fflib_IObjects
+{
+	protected List<Object> objects { get; private set;} { objects = new List<Object>(); }
+
+	/**
+	 * Class constructor
+	 */
+	public fflib_Objects(List<Object> objects)
+	{
+		this.objects = objects.clone();
+	}
+
+	public virtual Object getType()
+	{
+		return Object.class;
+	}
+
+	public List<Object> getObjects()
+	{
+		return this.objects;
+	}
+
+	public Boolean contains(Object value)
+	{
+		return getObjects()?.contains(value);
+	}
+
+	public Boolean containsAll(List<Object> values)
+	{
+		if (values == null) return false;
+
+		return containsAll(new Set<Object>(values));
+	}
+
+	public Boolean containsAll(Set<Object> values)
+	{
+		if (values == null) return false;
+
+		for (Object value : values)
+		{
+			if (!getObjects()?.contains(value)) return false;
+		}
+		return true;
+	}
+
+	public Boolean containsNot(Object value)
+	{
+		return !contains(value);
+	}
+
+	public Boolean containsNot(List<Object> values)
+	{
+		if (values == null) return true;
+
+		return containsNot(new Set<Object>(values));
+	}
+
+	public Boolean containsNot(Set<Object> values)
+	{
+		if (values == null) return true;
+
+		for (Object value : values)
+		{
+			if (getObjects()?.contains(value)) return false;
+		}
+		return true;
+	}
+
+	public Boolean isEmpty()
+	{
+		return (getObjects() == null || getObjects().isEmpty());
+	}
+
+	public Boolean isNotEmpty()
+	{
+		return !isEmpty();
+	}
+
+	protected void setObjects(List<Object> objects)
+	{
+		this.objects = objects;
+	}
+}
--- a/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls
@@ -39,12 +39,18 @@
  *
  **/
 public virtual with sharing class fflib_SObjectDomain
+		extends fflib_SObjects
 	implements fflib_ISObjectDomain
 {
 	/**
 	 * Provides access to the data represented by this domain class
 	 **/
-	public List<SObject> Records { get; private set;}
+	public List<SObject> Records {
+		get
+		{
+			return getRecords();
+		}
+	}
 
 
 	/**
@@ -74,19 +80,14 @@ public virtual with sharing class fflib_SObjectDomain
 	}
 
 	/**
-	 * Derived from the records provided during construction, provides the native describe for the standard or custom object
-	 **/
-	public Schema.DescribeSObjectResult SObjectDescribe {get; private set;}
-
-	/**
 	 * Exposes the configuration for this domain class instance
 	 **/ 
 	public Configuration Configuration {get; private set;}    
 		
 	/**
-	 * Useful during unit testign to assert at a more granular and robust level for errors raised during the various trigger events
-	 **/	
-	public static ErrorFactory Errors  {get; private set;}
+	 * DEPRECATED, This property has been moved to fflib_SObjects
+	 **/
+	public static fflib_SObjectDomain.ErrorFactory Errors {get; private set;}
 	
 	/**
 	 * Useful during unit testing to access mock support for database inserts and updates (testing without DML)
@@ -105,8 +106,8 @@ public virtual with sharing class fflib_SObjectDomain
 
 	static
 	{
-		Errors = new ErrorFactory();
-		
+		Errors = new fflib_SObjectDomain.ErrorFactory();
+
 		Test = new TestFactory();
 		
 		TriggerStateByClass = new Map<Type, List<fflib_SObjectDomain>>();
@@ -137,10 +138,9 @@ public virtual with sharing class fflib_SObjectDomain
 	public fflib_SObjectDomain(List<SObject> sObjectList, SObjectType sObjectType)
 	{
 		// Ensure the domain class has its own copy of the data
-		Records = sObjectList.clone(); 
-		// Capture SObjectType describe for this domain class
-		SObjectDescribe = sObjectType.getDescribe();
-		// Configure the Domain object instance 
+		super(sObjectList, sObjectType);
+
+		// Configure the Domain object instance
 		Configuration = new Configuration();		
 	}
 	
@@ -278,27 +278,11 @@ public virtual with sharing class fflib_SObjectDomain
     /**
      * Returns the SObjectType this Domain class represents
      **/
-    public SObjectType getSObjectType()
-    {
-    	return SObjectDescribe.getSObjectType();
-    }
-
-    /**
-     * Returns the SObjectType this Domain class represents
-     **/
     public SObjectType sObjectType()
     {
     	return getSObjectType();
     }
 
-    /**
-     * Alternative to the Records property, provided to support mocking of Domain classes
-     **/
-    public List<SObject> getRecords()
-    {
-    	return Records;
-    }
-    
 	/**
 	 * Detects whether any values in context records have changed for given fields as strings
 	 * Returns list of SObject records that have changes in the specified fields
@@ -480,7 +464,8 @@ public virtual with sharing class fflib_SObjectDomain
 		if(domains==null || domains.size()==0)
 			return null;		
 		fflib_SObjectDomain domain = domains.remove(domains.size()-1);
-		domain.Records = records;
+
+		domain.setObjects(records);
 		return domain;
 	}
 	
@@ -674,34 +659,34 @@ public virtual with sharing class fflib_SObjectDomain
 	/**
 	 * Ensures logging of errors in the Domain context for later assertions in tests
 	 **/
-	public String error(String message, SObject record)
+	public override String error(String message, SObject record)
 	{
-		return Errors.error(this, message, record);	
+		return fflib_SObjectDomain.Errors.error(this, message, record);
 	}
 	
 	/**
 	 * Ensures logging of errors in the Domain context for later assertions in tests
 	 **/
-	public String error(String message, SObject record, SObjectField field)
+	public override String error(String message, SObject record, SObjectField field)
 	{
-		return Errors.error(this, message, record, field);	
+		return fflib_SObjectDomain.Errors.error(this, message, record, field);
 	}
-	
+
 	/**
-	 * Ensures logging of errors in the Domain context for later assertions in tests
+	 * DEPRECATED, This class has been moved to fflib_SObjects
 	 **/
 	public class ErrorFactory
 	{
-		private List<Error> errorList = new List<Error>(); 
-		
+		private List<Error> errorList = new List<Error>();
+
 		private ErrorFactory()
 		{
-			
+
 		}
-		
+
 		public String error(String message, SObject record)
 		{
-			return error(null, message, record);	
+			return error(null, message, record);
 		}
 
 		private String error(fflib_SObjectDomain domain, String message, SObject record)
@@ -711,12 +696,12 @@ public virtual with sharing class fflib_SObjectDomain
 			objectError.message = message;
 			objectError.record = record;
 			errorList.add(objectError);
-			return message;	
+			return message;
 		}
-		
+
 		public String error(String message, SObject record, SObjectField field)
 		{
-			return error(null, message, record, field);	
+			return error(null, message, record, field);
 		}
 
 		private String error(fflib_SObjectDomain domain, String message, SObject record, SObjectField field)
@@ -727,48 +712,48 @@ public virtual with sharing class fflib_SObjectDomain
 			fieldError.record = record;
 			fieldError.field = field;
 			errorList.add(fieldError);
-			return message;	
-		}	
-			
+			return message;
+		}
+
 		public List<Error> getAll()
-		{ 
+		{
 			return errorList.clone();
 		}
-		
+
 		public void clearAll()
 		{
 			errorList.clear();
-		}					
+		}
 	}
 	
 	/**
-	 * Ensures logging of errors in the Domain context for later assertions in tests
+	 * DEPRECATED, This class has been moved to fflib_SObjects
 	 **/
 	public virtual class FieldError extends ObjectError
 	{
 		public SObjectField field;
-		
+
 		public FieldError()
-		{		
-			
+		{
+
 		}
 	}
-	
+
 	/**
-	 * Ensures logging of errors in the Domain context for later assertions in tests
+	 * DEPRECATED, This class has been moved to fflib_SObjects
 	 **/
 	public virtual class ObjectError extends Error
 	{
 		public SObject record;
-		
+
 		public ObjectError()
 		{
-		
+
 		}
 	}
-	
+
 	/**
-	 * Ensures logging of errors in the Domain context for later assertions in tests
+	 * DEPRECATED, This class has been moved to fflib_SObjects
 	 **/
 	public abstract class Error
 	{
--- a/sfdx-source/apex-common/main/classes/fflib_SObjects.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjects.cls
@@ -1,0 +1,368 @@
+/**
+ * Copyright (c), FinancialForce.com, inc
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ *   are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *      this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *      this list of conditions and the following disclaimer in the documentation
+ *      and/or other materials provided with the distribution.
+ * - Neither the name of the FinancialForce.com, inc nor the names of its contributors
+ *      may be used to endorse or promote products derived from this software without
+ *      specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ *  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ *  OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ *  THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ *  EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ *  OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ *  OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+**/
+public virtual class fflib_SObjects
+		extends fflib_Objects
+		implements fflib_ISObjects
+{
+
+	public Schema.DescribeSObjectResult SObjectDescribe {get; private set;}
+
+	/**
+	 * Useful during unit testing to assert at a more granular and robust level for errors raised during the various trigger events
+	 **/
+	public static ErrorFactory Errors  {get; private set;}
+
+	static
+	{
+		Errors = new ErrorFactory();
+	}
+
+	/**
+	 * Class constructor
+	 */
+	public fflib_SObjects(List<SObject> records)
+	{
+		super(records);
+	}
+
+	public fflib_SObjects(List<SObject> records, Schema.SObjectType sObjectType)
+	{
+		super(records);
+		SObjectDescribe = sObjectType.getDescribe();
+	}
+
+	public virtual List<SObject> getRecords()
+	{
+		return (List<SObject>) getObjects();
+	}
+
+	public virtual Set<Id> getRecordIds()
+	{
+		return new Map<Id, SObject>(getRecords()).keySet();
+	}
+
+	public virtual override Object getType()
+	{
+		return getSObjectType();
+	}
+
+	public virtual SObjectType getSObjectType()
+	{
+		return SObjectDescribe.getSObjectType();
+	}
+
+	/**
+	 * Ensures logging of errors in the Domain context for later assertions in tests
+	 *
+	 * @param message
+	 * @param record
+	 *
+	 * @return Returns the Error message
+	 **/
+	protected virtual String error(String message, SObject record)
+	{
+		return Errors.error(this, message, record);
+	}
+
+	/**
+	 * Ensures logging of errors in the Domain context for later assertions in tests
+	 *
+	 * @param message
+	 * @param record
+	 * @param field
+	 *
+	 * @return Returns the Error message
+	 **/
+	protected virtual String error(String message, SObject record, SObjectField field)
+	{
+		return fflib_SObjects.Errors.error(this, message, record, field);
+	}
+
+	protected virtual Set<Object> getFieldValues(Schema.SObjectField sObjectField)
+	{
+		Set<Object> result = new Set<Object>();
+		for (SObject record : getRecords())
+		{
+			result.add(record.get(sObjectField));
+		}
+		return result;
+	}
+
+	/**
+	 * @param sObjectField The Schema.SObjectField to compare against the given value
+	 * @param value The given value of the records sObjectField to include in the return
+	 *
+	 * @return A list with only the SObjects where the given sObjectField has the provided value
+	 */
+	protected virtual List<SObject> getRecordsByFieldValue(Schema.SObjectField sObjectField, Object value)
+	{
+		return getRecordsByFieldValues(sObjectField, new Set<Object>{value});
+	}
+
+	/**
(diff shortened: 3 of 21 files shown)


sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls
```diff
@@ -480,7 +464,10 @@ public virtual with sharing class fflib_SObjectDomain
 		if(domains==null || domains.size()==0)
 			return null;		
 		fflib_SObjectDomain domain = domains.remove(domains.size()-1);
-		domain.Records = records;
+
+		// todo - The next line is not compatible, as it changes the context of the domain at runtime,
+		// shouldn't we create a new instance of the domain with just those records?
+//		domain.Records = records;
```
wimvelzeboer, line 470: Any thoughts on how to best resolve this?

stohn777:
@wimvelzeboer

After deploying Mocks, pushing the PR branch results in ...
```
F:\salesforce_projects\fflib-apex-common>sfdx force:source:push
Job ID | 0Af1D000014uEPdSAM
SOURCE PROGRESS | █████████████████████████████████████░░░ | 34/37 Components
TYPE   PROJECT PATH                                                       PROBLEM
─────  ─────────────────────────────────────────────────────────────────  ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Error  sfdx-source\apex-common\test\classes\mocks\fflib_SObjectMocks.cls  Class fflib_SObjectMocks.SObjectDomain must implement the method: List<Object> fflib_Domain.getObjects() (31:23)
Error  sfdx-source\apex-common\test\classes\mocks\fflib_SObjectMocks.cls  Class fflib_SObjectMocks.SObjectDomain must implement the method: Object fflib_Domain.getType() (31:23)
Error  sfdx-source\apex-common\test\classes\fflib_ApplicationTest.cls     Dependent class is invalid and needs recompilation:
                                                                           Class fflib_SObjectMocks : Class fflib_SObjectMocks.SObjectDomain must implement the method: Object fflib_Domain.getType() (31:23)
ERROR running force:source:push:  Push failed.
```
cc. @ImJohnMDaniel

wimvelzeboer:
@stohn777 & @ImJohnMDaniel Thanks! I forgot to commit those. Now they should run successfully..

stohn777:
Hi @wimvelzeboer
I looked this over during the weekend and generally am satisfied with the change being backward compatible.

One question though.  The interface convention for the project is `fflib_IWidget`.  Is there a reason for the new Domain and DomainConstructor interfaces being named differently.

cc. @ImJohnMDaniel

wimvelzeboer:
@stohn777 Great to hear that you are satisfied with the proposed change!

About the naming convention, it is indeed true that in fflib the interface names start with an `I`. For a long time I was a huge fan of that, until I started working more and more with the same polymorphism principles that are used in force-di.  
There the names of the implementation is hardly used anywhere in the source code, the interface name is mainly used. Adding an `I` in front of the interface name than start to make less sense. 

I would find the following line better to read, than the line below that where we added all the `I`s
`Accounts accounts = (Accounts) Application.Domain.newInstance(records);`

`IAccounts accounts = (IAccounts) Application.Domain.newInstance(records);`

The implementation would then have the named something like  'AccountsImp', which is indeed not that nice to read. But since we hardly use that name (only in the Application or Force-Di configuration), that should not matter much.

Maybe its something for another discussion, but I leave that to you.

Cc: @ImJohnMDaniel @daveespo

wimvelzeboer:
@stohn777 @daveespo @ImJohnMDaniel @afawcett If you are interested in knowing what changes for fflib I am currently working on, then please have a look at my fork of this repo. https://github.com/wimvelzeboer/fflib-apex-common

There you can see the changes that I implemented so far and things that are on the road map to migrate to fflib.

daveespo (changes requested):
> Reviewed 2 of 7 files at r1, 3 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300)* status: 5 of 13 files reviewed, 6 unresolved discussions (waiting on @wimvelzeboer)

*[a discussion](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MRe8OaSAk_QcLBH8_Ev:-MRe8OaSAk_QcLBH8_Ew:bxj6oun) (no related file):*
@wimvelzeboer -- don't make any changes here yet -- still need to have some more discussion on this and I don't want to keep having you go back to rework the PR until all of the concerns are written down

Good work here; the separation of the Domain is about as clean as we could expect given the need to preserve compat

---
___
*[sfdx-source/apex-common/main/classes/fflib_Application.cls, line 334 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MRe2lH9Am-kyxVqy3U1:-MRe2lH9Am-kyxVqy3U2:b-heusdd) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/55a4272b829f78a6214180374f07f62e500095ef/sfdx-source/apex-common/main/classes/fflib_Application.cls#L334)):*
> ```VBA
> 		 *              or the constructor for Domain class was not registered for the SObjectType
> 		 **/
> 		public virtual fflib_Domain newInstance(List<SObject> records)
> ```

I'm not super well versed in fflib_Application -- but changing the return type on newInstance seems like it could break implementations that are depending on calling the `sObjectType()` method on the return value without casting it first. 

Are there realistic use cases where you wouldn't be casting the return value of newInstance to a concrete SObjectDomain implementation already?
___
*[sfdx-source/apex-common/main/classes/fflib_Domain.cls, line 26 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MRe2Huh1RzxgujYo4s-:-MRe2Huh1RzxgujYo4s0:b4ukx3e) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/55a4272b829f78a6214180374f07f62e500095ef/sfdx-source/apex-common/main/classes/fflib_Domain.cls#L26)):*
> ```VBA
>  *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
> **/
> public interface fflib_Domain
> ```

I see your reasoning in the comment thread with John Storey about why you omitted the "I" in the interface name -- but for the sake of consistency, we need to continue to use it. `fflib_IDomain`
___
*[sfdx-source/apex-common/main/classes/fflib_DomainConstructor.cls, line 26 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MRe2YzmD1L2XipWYoqm:-MRe2YzmD1L2XipWYoqn:brpvatu) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/55a4272b829f78a6214180374f07f62e500095ef/sfdx-source/apex-common/main/classes/fflib_DomainConstructor.cls#L26)):*
> ```VBA
>  *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
> **/
> public interface fflib_DomainConstructor
> ```

Same here .. need the "I" `fflib_IDomainConstructor`
___
*[sfdx-source/apex-common/main/classes/fflib_Objects.cls, line 28 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MRe3dU71ogOBbSp4OFv:-MRe3dU71ogOBbSp4OFw:b-aoxkhk) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/55a4272b829f78a6214180374f07f62e500095ef/sfdx-source/apex-common/main/classes/fflib_Objects.cls#L28)):*
> ```VBA
> public virtual class fflib_Objects implements fflib_Domain
> {
> 	protected List<Object> Objects { get; private set;}
> ```

lower-case O for `objects` -- I know that it's "Records" in SObjectDomain but that is non-conventional variable naming and we can rectify some of these code style inconsistencies across the project in a future effort -- let's not dig the hole any deeper here
___
*[sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls, line 470 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MQIKRmK9_X_gGJKgPZ--r1-470:-MRe6Q8S8Qwtoqt3gLr2:biur3gz) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/537a54516267950b013e31a6b838bc8223c1afc5/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls#L470)):*

Any thoughts on how to best resolve this?
</blockquote></details>

I don't understand how this is incompatible with the new implementation? You could declare a protected setter in the fflib_SObjects/fflib_Objects implementation and set the Records collection through that, no?

stohn777 (commented):
Going back to the "I" for Interface question.  Avoiding the debate of which is more readable than the other, the convention in this repo is with an "I" so my preference is to adhere to that.  Removing the "I" from existing code would break backward compatibility.

> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-:-MRgVHxM6KAfzXAOn124:b-wi09jt)* status: 5 of 13 files reviewed, 6 unresolved discussions (waiting on @wimvelzeboer)

stohn777 (commented):
> Reviewed 1 of 7 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300)* status: 7 of 13 files reviewed, 6 unresolved discussions (waiting on @daveespo and @wimvelzeboer)

___
*[sfdx-source/apex-common/main/classes/fflib_Application.cls, line 334 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MRe2lH9Am-kyxVqy3U1:-MRgX8ZY0Jt_BhM2CMKR:b-uwratd) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/55a4272b829f78a6214180374f07f62e500095ef/sfdx-source/apex-common/main/classes/fflib_Application.cls#L334)):*

I'm not super well versed in fflib_Application -- but changing the return type on newInstance seems like it could break implementations that are depending on calling the `sObjectType()` method on the return value without casting it first. 

Are there realistic use cases where you wouldn't be casting the return value of newInstance to a concrete SObjectDomain implementation already?
</blockquote></details>

Good eye, @daveespo, and I agree.  This would break backward compatibility with a compile-time exception for any existing code -- "Illegal assignment from fflib_Domain to fflib_ISObjectDomain".

And same for all of these `fflib_Domain newInstance(?)` methods.

Unfortunately we cannot overload based on return type.

**This is a no-go blocker for me.**
___
*[sfdx-source/apex-common/main/classes/fflib_Domain.cls, line 26 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MRe2Huh1RzxgujYo4s-:-MRpq6Te65b8OR38Fif8:ba31hd) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/55a4272b829f78a6214180374f07f62e500095ef/sfdx-source/apex-common/main/classes/fflib_Domain.cls#L26)):*

I see your reasoning in the comment thread with John Storey about why you omitted the "I" in the interface name -- but for the sake of consistency, we need to continue to use it. `fflib_IDomain`
</blockquote></details>

Agreed as previously commented.

stohn777 (changes requested):
Per previous comments.

wimvelzeboer:
@stohn777 @daveespo I think that all issues are addressed now. The following changes are done:
- Added 'i' to interface names
- Change 'Objects' into 'objects' in fflib_Objects
- added protected setObjects method to fix compatibility issues, @daveespo thanks for the advise!

About the change of return type. I do not think this will break backward compatibility that much. 

When I tried this change to one of my projects, one with 64K lines of code, it installed without any issues.
In most cases the return of the `newInstance()` method is immediately casted to the correct type, like:
`Accounts accountsDomain = **(Accounts)** Application.Domain.newInstance(accountRecords);`

There is only one scenario where you will require a very small change:
`fflib_ISObjectDomain domain = Application.Domain.newInstance(records);`
would become:
`fflib_ISObjectDomain domain = **(fflib_ISObjectDomain)** Application.Domain.newInstance(records);`

To put this backwards compatibility issue in perspective; this change is smaller than the recent changes (added methods) in the fflib_UnitOfWork.IDML interface which also broke backwards compatibility.

sfdx-source/apex-common/main/classes/fflib_SObjects.cls
```diff
@@ -0,0 +1,131 @@
...
+		return result;
+	}
+
+	public virtual Schema.SObjectType getObjectType()
+	{
+		return SObjectDescribe.getSObjectType();
+	}
+
+	protected virtual fflib_SObjects selectByFieldValue(Schema.SObjectField sObjectField, Object value)
+	{
```
ImJohnMDaniel, line 84: Please a NullCheck on at least the the `value` parameter.

sfdx-source/apex-common/main/classes/fflib_SObjects.cls
```diff
@@ -0,0 +1,131 @@
...
+				new Set<Object>{ value }
+		);
+	}
+
+	protected virtual fflib_SObjects selectByFieldValues(Schema.SObjectField sObjectField, Set<Object> values)
+	{
+		List<SObject> result = new List<SObject>();
+		for (SObject record : getRecords())
+		{
+			if (values.contains(record.get(sObjectField)))
```
ImJohnMDaniel, line 96: Please add a safe navigation operator here between the "values" and the "contains method"

sfdx-source/apex-common/main/classes/fflib_SObjects.cls
```diff
@@ -0,0 +1,131 @@
...
+	 */
+	protected virtual void setFieldValueByMap(
+			Schema.SObjectField sObjectFieldToCheck,
+			Schema.SObjectField sObjectFieldToUpdate,
+			Map<Object, Object> values)
+	{
+		for (SObject record : getRecords())
+		{
+			Object keyValue = record.get(sObjectFieldToCheck);
+			if (values.containsKey(keyValue))
```
ImJohnMDaniel, line 125: please add a safe navigation operator between the `values` parameter and the `containsKey` method.

sfdx-source/apex-common/main/classes/fflib_SObjects.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
ImJohnMDaniel, line 3: v51.0 please

ImJohnMDaniel (changes requested):
@wimvelzeboer -- Thanks again for these changes.

I requested some changes mainly focused around null checks to avoid some NPEs.  

I am also requesting that all new classes be bumped to API v51.0.  We have another effort underway right now to bump all API versions to v51.0 and I would prefer it if these new classes could also start out at 51.0.

Seriously though -- I really like this changes!  We owe you a beer!

daveespo (commented):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300)* status: 3 of 17 files reviewed, 21 unresolved discussions (waiting on @daveespo and @wimvelzeboer)

___
*[sfdx-source/apex-common/main/classes/fflib_Application.cls, line 334 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MRe2lH9Am-kyxVqy3U1:-MUp0qrj2Fh5h5g9PkBS:b529wn3) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/55a4272b829f78a6214180374f07f62e500095ef/sfdx-source/apex-common/main/classes/fflib_Application.cls#L334)):*

Good eye, @daveespo, and I agree.  This would break backward compatibility with a compile-time exception for any existing code -- "Illegal assignment from fflib_Domain to fflib_ISObjectDomain".

And same for all of these `fflib_Domain newInstance(?)` methods.

Unfortunately we cannot overload based on return type.

**This is a no-go blocker for me.**
</blockquote></details>

After some discussion with John Storey, we agreed that this particular backwards incompatible change is fairly trivial to resolve -- this is no longer considered a blocker

sfdx-source/apex-common/main/classes/fflib_Objects.cls
```diff
@@ -25,13 +25,15 @@
...
 {
-	protected List<Object> objects { get; private set;}
+	protected List<Object> objects { get; private set;} { objects = new List<Object>(); }
 
 	/**
 	 * Class constructor
 	 */
 	public fflib_Objects(List<Object> objects)
 	{
+		if (objects == null) throw new DeveloperException('A domain cannot be constructed with a null value');
```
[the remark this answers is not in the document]
wimvelzeboer: Yeah, it makes more sense to translate a null into an empty list.

daveespo (changes requested):
> Reviewed 7 of 14 files at r4, 7 of 8 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300)* status: 16 of 17 files reviewed, 25 unresolved discussions (waiting on @ImJohnMDaniel, @stohn777, and @wimvelzeboer)

___
*[sfdx-source/apex-common/main/classes/fflib_Objects.cls, line 50 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MQIKRmK9_X_gGJKgPZ2-r4-50:-MUytnntDqxkkEPeDlAB:bsfr86a) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/bba78fb4365425597372223392e1853b2f757ec1/sfdx-source/apex-common/main/classes/fflib_Objects.cls#L50)):*
<details><summary><i>Previously, ImJohnMDaniel (John M. Daniel) wrote…</i></summary><blockquote>

Along with the overall request being made to bump the API version to v51.0, I would like to request the usage of the new "safe navigation operator".

```
return getObjects()?.contains(value);
```
</blockquote></details>

I don't agree with this change -- we've added a static initializer for the `objects` member -- it should never be null -- if it is, it's due to an underlying flaw in someone's implementation of an extension class (i.e. assigning it to null) .. so returning null from this method (which is what will happen) is unexpected (it should be a Boolean) and make for a truly confusing debugging session

I think just allowing the NullPointerException to be raised is the best developer experience 
___
*[sfdx-source/apex-common/main/classes/fflib_Objects.cls, line 62 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MQIKRmK9_X_gGJKgPZ2-r4-62:-MUyvUgS7pm9w5Pp_unC:bbru8a4) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/bba78fb4365425597372223392e1853b2f757ec1/sfdx-source/apex-common/main/classes/fflib_Objects.cls#L62)):*
<details><summary><i>Previously, ImJohnMDaniel (John M. Daniel) wrote…</i></summary><blockquote>

Please add a safe navigation operator here.
</blockquote></details>

Same as above; I don't think the safe nav operator is appropriate
___
*[sfdx-source/apex-common/main/classes/fflib_Objects.cls, line 81 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MQIKRmK9_X_gGJKgPZ2-r4-81:-MUyw6J4C0Pv44P2nVWI:b-cxba3s) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/bba78fb4365425597372223392e1853b2f757ec1/sfdx-source/apex-common/main/classes/fflib_Objects.cls#L81)):*
<details><summary><i>Previously, ImJohnMDaniel (John M. Daniel) wrote…</i></summary><blockquote>

Please add a safe navigation operator here.
</blockquote></details>

Same as above; safe navigation isn't appropriate
___
*[sfdx-source/apex-common/main/classes/fflib_Objects.cls, line 35 at r5](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MQIKRmK9_X_gGJKgPZ2-r5-35:-MUyxIwO7ENfgN8dmr_P:b-r480d1) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/e56beb06362a878bf4bdea6bc2de491f7cd4a567/sfdx-source/apex-common/main/classes/fflib_Objects.cls#L35)):*

Yeah, it makes more sense to translate a null into an empty list.
</blockquote></details>

Since this constructor is trying to replicate expected behavior of the old `fflib_SObjectDomain`, it needs to clone and I think we should allow the NullPointerException to be thrown if a null collection is passed in -- again, this will clearly describe the root of the developer error rather than masking it by creating an empty collection

If we want to add a convenience no-arg constructor that will construct an empty `objects` collection, sure fine .. but it seems like an fflib_Objects instance is pretty useless if it's got an empty `objects` collection so I don't think it's a real world case that you'd want to have it be empty, ever
___
*[sfdx-source/apex-common/main/classes/fflib_Objects.cls, line 81 at r6](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MUyvoB285plymiDuPSY:-MUyvoB285plymiDuPSZ:b3e2p0i) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/ab989a053281cb442ceceae4f57c1fd04dd07503/sfdx-source/apex-common/main/classes/fflib_Objects.cls#L81)):*
> ```VBA
> 	public Boolean containsNot(List<Object> values)
> 	{
> 		if (values == null) return false;
> ```

I think this is the incorrect return value -- for "containsNot", it should return True if you pass in a null `values` collection since that's what it would return if you passed in an empty collection
___
*[sfdx-source/apex-common/main/classes/fflib_Objects.cls, line 88 at r6](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MUyw3uB3Ix7UduKIQ2T:-MUyw3uB3Ix7UduKIQ2U:b-v3qmdi) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/ab989a053281cb442ceceae4f57c1fd04dd07503/sfdx-source/apex-common/main/classes/fflib_Objects.cls#L88)):*
> ```VBA
> 	public Boolean containsNot(Set<Object> values)
> 	{
> 		if (values == null) return false;
> ```

Same as above; should return true
___
*[sfdx-source/apex-common/main/classes/fflib_SObjects.cls, line 125 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MQIKRmK9_X_gGJKgPZ4-r4-125:-MUz9oGR78fZ29vgR0uS:b-d1yvf4) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/bba78fb4365425597372223392e1853b2f757ec1/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L125)):*
<details><summary><i>Previously, ImJohnMDaniel (John M. Daniel) wrote…</i></summary><blockquote>

please add a safe navigation operator between the `values` parameter and the `containsKey` method.
</blockquote></details>

I don't like this change for the same reasons outlined elsewhere; you should never pass in a null `values` Map .. it masks developer errors behind code that appears to work
___
*[sfdx-source/apex-common/main/classes/fflib_SObjects.cls, line 49 at r6](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MUz5CTm7xwWFjWl8zbX:-MUz5CTnDkKEZjVosu3A:bg2n7t4) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/ab989a053281cb442ceceae4f57c1fd04dd07503/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L49)):*
> ```VBA
> 	{
> 		super(records);
> 		if (sObjectType == null && records.getSObjectType() != null)
> ```

I am looking at the old `fflib_SObjectDomain` constructor and it didn't have this behavior of falling back to the sObjectType of the List -- but that code isn't even really reachable because an exception will be raised in the else/if block 5 lines down if `sObjectType` is null

I propose that we keep the old behavior of requiring both the `records` and `sObjectType` to be non-null 

If we'd like to be helpful to the developer, we can `System.assert(records != null, 'records collection cannot be null')` but then we are introducing more error messages that aren't localized so ultimately I think I'd prefer to just let the NullPointerException get raised 
___
*[sfdx-source/apex-common/main/classes/fflib_SObjects.cls, line 78 at r6](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MUz7YfNAEg99LExWZ_V:-MUz7YfNAEg99LExWZ_W:b-6l5qpo) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/ab989a053281cb442ceceae4f57c1fd04dd07503/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L78)):*
> ```VBA
> 	}
> 
> 	protected virtual Set<Object> getFieldValues(Schema.SObjectField sObjectField)
> ```

This is a protected method that isn't called from anywhere in fflib -- additionally, I'm not sure what I'd expect the behavior to be (should it return a Set or a List, possibly containing duplicates)

For this reason, I think we should remove the implementation from the core library 
___
*[sfdx-source/apex-common/main/classes/fflib_SObjects.cls, line 271 at r6](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300#-MUzA2vPDb6XrgntKSlc:-MUzA2vQ1dB4TZsTT71n:b80iqex) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/ab989a053281cb442ceceae4f57c1fd04dd07503/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L271)):*
> ```VBA
> 	{
> 		if (!(objects instanceof List<SObject>))
> 			throw new DeveloperException('fflib_SObjects can only accept an instance of List<SObject>');
> ```

I'd like to `System.assert()` here rather than invent our own new Exception

wimvelzeboer:
> > sfdx-source/apex-common/main/classes/fflib_SObjects.cls, line 78 at r6 (raw file):
> > protected virtual Set<Object> getFieldValues(Schema.SObjectField sObjectField)
> This is a protected method that isn't called from anywhere in fflib -- additionally, I'm not sure what I'd expect the behavior to be > (should it return a Set or a List, possibly containing duplicates)
> For this reason, I think we should remove the implementation from the core library

The idea behind this method is to make it available for just only the domain classes. 
Domain should be aware in which fields data is being stored, a service using the domain should not. That's why the protected.
Many methods in the domain are about getting data from fields and returning those values.
A method on a domain could look like the following, where you just call the method instead of having iterations in every getter method:
`
public class AccountsImp extends fflib_sObjects implements IAccounts 
{
  public Set<String> getShippingCountry()
  {
    return (Set<String>) getFieldValues(Schema.Account.ShippingCountry);
  }
}
`

sfdx-source/apex-common/test/classes/fflib_ObjectsTest.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
PeterLinBiz, line 3: @wimvelzeboer Could you please update this to 51.0? Thanks!

sfdx-source/apex-common/test/classes/fflib_SObjectsTest.cls-meta.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>48.0</apiVersion>
```
PeterLinBiz, line 3: @wimvelzeboer Could you please update this to 51.0? Thanks!

stohn777 (dismissed):

daveespo (commented):
> Reviewed 1 of 2 files at r7, 2 of 6 files at r8, 2 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300)* status: 19 of 21 files reviewed, 23 unresolved discussions (waiting on @daveespo, @ImJohnMDaniel, @stohn777, and @wimvelzeboer)

stohn777:
@wimvelzeboer
PR-300's branch is behind Master.  Would you please merge therefrom and resolve conflicts as needed, which I believe will alleviate many of the test failures?

wimvelzeboer:
@daveespo @ImJohnMDaniel @stohn777 I am currently working on updating the PR for apex-common-examples, and found a few tiny things that were missing in this PR. I will keep adding them here. Before you merge this PR please let me know, then I want to rebase this branch again and squash all the commits here into a single one.

daveespo:
@wimvelzeboer  -- please don't add any new functionality to this PR; bug fixes or code review fixes only

It's a bear of a PR and all of us have reviewed it multiple times -- we don't want to have to go back and do it again :-)

wimvelzeboer:
> @wimvelzeboer -- please don't add any new functionality to this PR; bug fixes or code review fixes only
> 
> It's a bear of a PR and all of us have reviewed it multiple times -- we don't want to have to go back and do it again :-)

yeah, I am only fixing some issues with things I forgot to move between fflib_SObjectDomain and fflib_SObject. Like the error class. And I want to double check the ApexDocs for the methods (where it makes sense)

PeterLinBiz:
@wimvelzeboer 
I got the same 4 errors while I was trying to create an Unlocked Package version as this "Checks" run at https://github.com/apex-enterprise-patterns/fflib-apex-common/pull/300/checks .
=== Errors
(1) Apex Test Failure: Class.fflib_SObjectDomainTest.testInsertValidationFailedWithoutDML: line 55, column 1 System.TypeException: Invalid conversion from runtime type fflib_SObjects.FieldError to fflib_SObjectDomain.FieldError
(2) Apex Test Failure: Class.fflib_SObjectDomainTest.testErrorLogging: line 154, column 1 System.TypeException: Invalid conversion from runtime type fflib_SObjects.FieldError to fflib_SObjectDomain.FieldError
(3) Apex Test Failure: Class.fflib_SObjectDomainTest.testValidationWithoutDML: line 37, column 1 System.TypeException: Invalid conversion from runtime type fflib_SObjects.FieldError to fflib_SObjectDomain.FieldError
(4) Apex Test Failure: Class.fflib_SObjectDomainTest.testUpdateValidationFailedWithoutDML: line 73, column 1 System.TypeException: Invalid conversion from runtime type fflib_SObjects.FieldError to fflib_SObjectDomain.FieldError

The strange was I ran this same Class.fflib_SObjectDomainTest right before I created the version but all classes and all test methods passed in the Illuminated Cloud (IC). 
After I got errors from creating version.
IC started to show me the same 4 errors above.
The Git commit ID was dab9a872cf9136f0928bfa8e5688badf9d82f34f
Illuminated Cloud: 2.1.6.6
WebStorm: WebStorm 2020.3.2 Build #WS-203.7148.54, built on January 25, 2021.

PeterLinBiz:
@wimvelzeboer 
Sorry, when I wrote the above "The strange was I ran this same Class.fflib_SObjectDomainTest right before I created the version but all classes and all test methods passed in the Illuminated Cloud (IC)" I probably ran all test on the commit before the commit above.
That commit ID was: 840d65ffcb27e5cf7affadfdd4b4ecf09cb4f785 .
I just created a new scratch org with the above commit ID and all tests passed.

stohn777:
Hi @wimvelzeboer
For the checks failure from the latest commit and from a quick browse, I suspect `fflib_SObjectDomain` line 90 should be:
`public static fflib_SObjectDomain.ErrorFactory Errors  = fflib_SObjectDomain.Errors;`

Also I see that your fork's master is still behind AEP's master, but that only reflects the recent API Version upgrades and would affect the test.

PeterLinBiz:
@wimvelzeboer 
I have packaged the Commit ID 52fd335 as an Unlocked Package Version and installed in my sandboxes, I ran my test classes, and all passed.
Thanks for this new Commit!
It is working great!

wimvelzeboer:
Hi @stohn777 @daveespo @ImJohnMDaniel @afawcett 
Is there still anything pending before this PR can get merged?
If there is please let me know!

ImJohnMDaniel (approved):
> Reviewed 1 of 1 files at r3, 4 of 14 files at r4, 6 of 8 files at r5, 1 of 2 files at r7, 2 of 6 files at r8, 2 of 2 files at r9, 1 of 2 files at r10, 4 of 4 files at r11.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/300)* status: all files reviewed, 23 unresolved discussions (waiting on @ImJohnMDaniel, @stohn777, and @wimvelzeboer)

stohn777 (approved):

PR #310, opened 2020-12-31
Add support for domain in UnitOfWork

wimvelzeboer:
With this change you can directly pass a domain class to the unitOfWork to process the records of the domain instead of first have to extract the records and then pass them to the domain\

`unitOfWork.registerNew(accounts.getRecords());` then becomes `unitOfWork.registerNew(accounts);`.

Small change but will make things easier to read.

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/310)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_ISObjectUnitOfWork.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_ISObjectUnitOfWork.cls
@@ -42,6 +42,12 @@ public interface fflib_ISObjectUnitOfWork
      **/
     void registerNew(List<SObject> records);
     /**
+     * Register a list of newly created SObject instances contained in a domain to be inserted when commitWork is called
+     *
+     * @param domain A domain containing newly created SObject instances to be inserted during commitWork
+     **/
+    void registerNew(fflib_ISObjects domain);
+    /**
      * Register a newly created SObject instance to be inserted when commitWork is called,
      *   you may also provide a reference to the parent record instance (should also be registered as new separately)
      *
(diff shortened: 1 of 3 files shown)


stohn777 (dismissed):

wimvelzeboer:
@stohn777 Do you mean that I have used a mix in indentation type; space/tab? I tried to use the same indentation type as used in the file. There is currently a mix between space and tab, which one is preferred? Maybe an idea to have one commit that fixes this inconsistency.

JAertgeerts (approved):

wimvelzeboer:
@stohn777 I also rebased this branch and updated it to the new fflib_ISObjects interface structure

wimvelzeboer:
@stohn777 I have fixed the identation!

wimvelzeboer:
I also come up with an idea to keep track of changed fields in the Domain, and that when you pass the domain to the UnitOfWork it automatically extracts those changes fields and marks them as dirty. Similar to the method `public void registerDirty(SObject record, List<SObjectField> dirtyFields)`. 

@afawcett @daveespo @stohn777 @ImJohnMDaniel What are your thoughts on that features? 
I could add that to this PR, but I would not want to hold this PR for too long.

stohn777:
@wimvelzeboer 
During our last PR collaboration meeting at the end of May, we agreed to decline this PR for two reasons:

1. Whereas Selectors and UOWs are generally consumed by Services and Domains, this PR inverts that hierarchy by making UOW loosely dependent on Domain.  This being the primary concern.
2. Additionally as you alluded to in the initial commit, this change provides a very minor benefit.

My apology for not getting the decision out sooner resulting in some extra effort on your part.

wimvelzeboer:
@stohn777 I do not fully agree with that reasoning. In my opinion the UnitOfWork should be used primarily by the Service and Trigger Handler (onAfter*) and should never appear on any Domain classes. If it does, then that would tie the data directly to a source. 

For example; if I have a process "A" on the domain changing some data, and it would be directly tied to the UnitOfWork. Then I cannot use that same process "A" again for another data source (e.g. Platform Cache, External WebService Push). In that case I have to duplicate process "A" and name it "B". 

If the domain is completely separated from the data source, then a multiple service methods can invoke the same data changing process "A" and then decide where the data should be stored. 

There is a perfect example if this issue of conflict in the fflib-apex-common-samplecode. The `applyDiscount` method in the [Opportunities Domain](https://github.com/apex-enterprise-patterns/fflib-apex-common-samplecode/blob/master/sfdx-source/apex-common-samplecode/main/classes/domains/Opportunities.cls) requires an unitOfWork and directly registers the changed data to its data source. 
This direct reference between data and source is very limiting, what if those records originated from e.g. the Platform Cache?

That is why you would want to turn the logic around, and have the service decide where the data should go, and then it would make sense to send the Domain to the UnitOfWork. Where the Domain can tell the UnitOfWork which fields has been changed.

PR #250, opened 2019-12-02
Feature/fluent selector constructor

wimvelzeboer:
This will add a more fluent design to the fflib_SObjectSelector class constructor, so that you can configure it without having to use the Boolean constructor paramaters.

`AccountsSelector selector = new AccountsSelector().enforceFLS();`

Added methods are:
`
enforceFLS();
includeFieldSetFields();
ignoreCRUD();
unsortedSelectFields();
`

PR #311, opened 2021-01-07
More fluent design for selector

wimvelzeboer:
At my work I noticed that a lot of selector methods were duplicated, the only difference between them were the offset, limit and ordering.
Then I thought, it would be nice to have one selector method and that the method invoking that selector could decide about the ordering, offset and limit, without having to pass those requirements as arguments to the selector method.
That the selector can be called via something like:
`List<Account> records = new AccountsSelector().setOffset(10).selectById(idSet);`

This change is implementing that idea.

Here is an example on how that would look like:
AccountsSelector

```
public with sharing class AccountsSelector extends fflib_SObjectSelector
{
    ...
    public static fflib_QueryFactory.Ordering largeAccountsFirst =
            new fflib_QueryFactory.Ordering(Account.NumberOfEmployees, fflib_QueryFactory.SortOrder.DESCENDING);

    public static AccountsSelector newInstance()
    {
        return ((AccountsSelector) Application.Selector.newInstance(Schema.Account.SObjectType))
    }

    public List<Account> selectById(Set<Id> idSet)
    {
        return (List<Account>) selectSObjectsById(idSet);
    }
    ....
}
```

AccountsService class
```
public with sharing class AccountsService
{
    public void MyServiceMethod(Set<Id> idSet)
    {
        List<Account> records = 
                AccountsSelector.newInstance()
                        .setOrdering(AccountsSelector.largeAccountsFirst)
                        .setLimit(10)
                        .selectById(idSet);
        ....
    }
}
```

This would avoid selector method like this one:
`public List<Case> selectClosedCasesByContact(String orderQuery, Set<Id> contactIds, Integer limitSize, Integer
offSet);`

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/311)


JAertgeerts (approved):

wimvelzeboer:
Hi @ImJohnMDaniel @stohn777 @afawcett @daveespo,
What are your thoughts on this proposal for the selector?

ImJohnMDaniel:
G'day @wimvelzeboer, this topic came up on our monthly call yesterday.

Question for you.  Besides the following method names, what else would we want to add to the fluent methods?
- setOffset
- setLimit
- setOrdering

wimvelzeboer:
Hi @ImJohnMDaniel,

This PR will add the following methods for a more fluent selector:
- setOrdering
- setLimit
- setOffset

Those methods are an addition to the other methods added in a similar PR:
https://github.com/apex-enterprise-patterns/fflib-apex-common/pull/250
Those methods where:
- enforceFLS
- includeFieldSetFields
- ignoreCRUD
- unsortedSelectFields

There is one thing I would like to discuss. The apex-common-examples create selector instances in the following way:
```
List<Account> records = AccountsSelector.newInstance().selectById(ids);
```
Using the following static method:
```
public static IAccountsSelector newInstance()
{
    return ((IAccountsSelector) Application.Selector.newInstance(Schema.Account.SObjectType))
}
```
You can't use the fluent methods with this approach, as the method returns an Interface reference. And the fluent methods are not defined in that.
The fluent methods only work when the `newInstance()` method returns the implementation `AccountsSelector`.

Would if be an idea to add the fluent methods to the interface fflib_ISObjectsSelector, potentially breaking backwards compatibility. I think the impact will be similar to the changes to the fflib_SObjectUnitOfWork.IDML interface.

What do you think? 

Ps. If you want me to have a call again about the open PRs to discuss them, just let me know.

wimvelzeboer:
@ImJohnMDaniel We could maybe add a method to add additional fields, other then the standard list returned by `getSObjectFieldList()`. 
I do often see `selectByIdWithXXX` methods which are basically only a selectById and adding some rare fields that one specific process is using.

JAertgeerts:
Hi @ImJohnMDaniel, can we merge this?

daveespo:
@wimvelzeboer  and @JAertgeerts  -- we discussed this PR some more and are likely to reject it. 

We don't believe changing the largely stateless Selector model into a more stateful `SObjectSelector`. The legacy `enforceFLS` and `enforceCRUD` pieces of state have been part of fflib since its creation so it wasn't controversial to make them settable via fluent methods. But this PR proposes moving a lot of state up from the `QueryFactory` into `SObjectSelector`.

Would you be supportive of making more methods in `QueryFactory` fluent to support your use cases?

wimvelzeboer:
> Would you be supportive of making more methods in QueryFactory fluent to support your use cases?

Well, the reason of this change is that I see a lot of duplication in the selector method, which is mainly just about a different ordering, or limits. Things that are determined based on the business logic which is invoking those selector methods. 

The QueryFactory is more hidden behind the selector implementation and cannot directly be reached from the business logic (e.g. service class).

Now I see a lot of these kind of methods:
```
public List<Account> selectById(Set<Id> idSet)
{
	return (List<Account>) selectSObjectsById(idSet);
}
public List<Account> selectById(Set<Id> idSet, Integer limit, Integer offset)
{
	return (List<Account>) Database.query(
				newQueryFactory()
					.setCondition('Id = :idSet')
					.setLimit(limit)
					.setOffset(offset)
					.toSOQL());
}
public List<Account> selectById(Set<Id> idSet, fflib_QueryFactory.Ordering ordering)
{
	return (List<Account>) Database.query(
				newQueryFactory()
					.setCondition('Id = :idSet')
					.setOrdering(ordering)
					.toSOQL());
}
public List<Account> selectById(Set<Id> idSet,  Integer limit, Integer offset, fflib_QueryFactory.Ordering ordering)
{
	return (List<Account>) Database.query(
				newQueryFactory()
					.setCondition('Id = :idSet')
					.setOrdering(ordering)
					.setLimit(limit)
					.setOffset(offset)
					.toSOQL());
}
``` 
@daveespo How would you then exactly propose to limit these duplication, without using a fluent design, and using a more fluent QueryFactory like you propose?

daveespo:
@wimvelzeboer  what you propose is what I exactly expect -- i.e. the fluent methods are within QueryFactory for `setOrdering`, `setLimit`, `setOffset` and `setCondition` (and those already exist today)

Because one of the documented common usage patterns of Selectors presumes reuse of the Selector instance, we see that the risk of this PR outweighs the benefit. Users would be surprised to find out that the second query result is limited to 10 records.

```
AccountSelector aSel = new AccountSelector();

... several lines of business logic ...

List<Account> daveList = aSel.setLimit(10).selectByName('David');

... several lines of business logic ...

List<Account> usAccounts = aSel.selectByCountry('US');

```

We acknowledge that in your sample code on the first comment on this PR that you would not face a problem since you're constructing a new instance of the Selector each time you use it ... whether that's a best practice or anti-pattern is a debate that we aren't interested in having ... but given my example above, you could see how adding state to Selector that never existed before violates the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

PR #313, opened 2021-01-11
A common criteria based filter for Domains and Selectors

wimvelzeboer:
A common criteria based filter for domains and SOQL condition generator for Selectors.

Often the same filters are repeated in the domain and the selector classes. The Criteria feature provides a solution for that by extracting the filter conditions into a single reusable criteria class. The filter conditions are made dynamic so that they can be evaluated in run-time, or be converted to a SOQL statement condition.
```
                + - - - - - - - - +
        + - - - | Filter Criteria | - - - +  
        |       + - - - - - - - - +       |
        |                                 | 
        |                                 |
+ - - - - - - - +                 + - - - - - - - +
|    Domain     |                 |    Selector   |
+ - - - - - - - +                 + - - - - - - - +
```
Here is an example on how its used:

The criteria class is the place where all the filter conditions are stored for a single SObjectType.

```apex
public with sharing class AccountCriteria extends fflib_Criteria
{
    public AccountCriteria ShippingCountryEquals(String countryName)
    { 
        equalTo(Schema.Account.ShippingCountry, countryName);
        return this;                
    }
    
    public AccountCriteria NumberOfEmployeesGreaterThan(Integer numberOfEmployees)
    {
        greaterThan(Schema.Account.NumberOfEmployees, numberOfEmployees);
        return this;
    }
}
```

How it can be applied in a Domain class:

```apex
public with sharing class Accounts
        extends fflib_SObjectDomain
        implements IAccounts
{
    private static final Integer LARGE_ACCOUNT_EMPLOYEE_NUMBERS = 500;

    public Accounts getByCountry(String countryName)
    {
        return new Accounts(
                getRecords(
                        new AccountCriteria().ShippingCountryEquals(countryName)
                )
        );
    }
    
    public Accounts getByNumberOfEmployeesGreaterThan(Integer numberOfEmployees)
    {
        return new Accounts(
                getRecords(
                       new AccountCriteria().NumberOfEmployeesGreaterThan(numberOfEmployees)
                )
        );
    }
    
    public Accounts getByLargeAccountsInCountry(String countryName)
    {
        return new Accounts(
                getRecords(
                       new AccountCriteria()
                               .ShippingCountryEquals(countryName)
                               .NumberOfEmployeesGreaterThan(numberOfEmployees)
                )
        );
    }
}
```
In this example we see three filters; one for country, another for checking minimal number of employees and a third that combines the first two.
It is important not to have a filter with too many conditions. 
One filter criteria condition per method is ideal to have maximum flexibility and a high chance on code-reuse.

How the same filters can be used in the Selector class:

```apex
public with sharing class AccountsSelector
        extends fflib_SObjectSelector
        implements IAccountsSelector
{
    ...
    public List<Account> selectByCountryWithMinimalNumberOfEmployees(String country, Integer minimalNumberOfEmployees)
    {
        return (List<Account>)  Database.query(
                newQueryFactory()
                        .setCondition(
                                new AccountCriteria()
                                        .ShippingCountryEquals(country)
                                        .NumberOfEmployeesGreaterThan(minimalNumberOfEmployees)
        );
    }
    ...
}
```

With this change developers can avoid a lot of code duplications.
Hope you like it!

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/313)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_Criteria.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_Criteria.cls
@@ -1,0 +1,654 @@
+/**
+ * Copyright (c), FinancialForce.com, inc
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ *   are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *      this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *      this list of conditions and the following disclaimer in the documentation
+ *      and/or other materials provided with the distribution.
+ * - Neither the name of the FinancialForce.com, inc nor the names of its contributors
+ *      may be used to endorse or promote products derived from this software without
+ *      specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ *  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ *  OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+ *  THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ *  EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ *  OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ *  OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+**/
+public virtual with sharing class fflib_Criteria
+		implements Evaluator
+{
+	// Add brace when generated to SQL
+	private Boolean embraced = false;
+
+	// List to hold all evaluators
+	private List<Evaluator> evaluators;
+
+	// The type of comparator NONE | AND | OR
+	private String comparatorType;
+
+
+	/**
+	 * populate private variables with default settings.
+	 */
+	public fflib_Criteria()
+	{
+		this.evaluators = new List<Evaluator>();
+		this.comparatorType = 'AND';
+	}
+
+	/**
+	 * Changes the default comparator for each criteria to OR
+	 *
+	 * @return Instance of fflib_Criteria
+	 *
+	 * @example
+	 *  new fflib_Criteria()
+	 *          .orCriteria()
+	 *          .equalTo(Account.Name, 'Example')
+	 *  		.equalTo(Account.AccountNumber, '1234567')
+	 *
+	 *  Evaluates:
+	 *      Name = 'Example' OR AccountNumber = '1234567'
+	 */
+	public fflib_Criteria orCriteria()
+	{
+		this.comparatorType = 'OR';
+		return this;
+	}
+
+	/**
+	 * Changes the default comparator for each criteria to AND.
+	 * By default the comparator is set to AND,
+	 * so this method should only be used in custom implementations extended from fflib_Criteria
+	 *
+	 * @return Instance of fflib_Criteria
+	 *
+	 * @example
+	 *  new MyCustomCriteria()
+	 *          .andCriteria()
+	 *          .equalTo(Account.Name, 'Example')
+	 *  		.equalTo(Account.AccountNumber, '1234567')
+	 *
+	 *  Evaluates:
+	 *      Name = 'Example' AND AccountNumber = '1234567'
+	 */
+	public fflib_Criteria andCriteria()
+	{
+		this.comparatorType = 'AND';
+		return this;
+	}
+
+	/**
+	 * Adds a sub criteria with OR comparator
+	 *
+	 * @param subCriteria The condition of the sub criteria
+	 *
+	 * @return Instance of fflib_Criteria
+	 *
+	 * @example
+	 * new fflib_Criteria()
+	 *          .equalTo(Account.Name, 'Example')
+	 *          .addOrCriteria(
+	 *              new fflib_Criteria()
+	 *              		.equalTo(Account.AccountNumber, '0001')
+	 *              		.equalTo(Account.AccountNumber, '0002'))
+	 *
+	 *  Evaluates:
+	 *      Account.Name = 'Example' AND (Account.AccountNumber = '0001' OR Account.AccountNumber = '0002')
+	 */
+	public fflib_Criteria addOrCriteria(fflib_Criteria subCriteria)
+	{
+		subCriteria.orCriteria();
+		subCriteria.setEmbraced(true);
+		evaluators.add(subCriteria);
+		return this;
+	}
+
+	/**
+	 * Adds a sub criteria with AND comparator
+	 *
+	 * @param subCriteria The condition of the sub criteria
+	 *
+	 * @return Instance of fflib_Criteria
+	 *
+	 * @example
+	 * new fflib_Criteria()
+	 *          .orCriteria()
+	 *          .equalTo(Account.Name, 'Example')
+	 *          .addAndCriteria(
+	 *              new fflib_Criteria()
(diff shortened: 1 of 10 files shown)


sfdx-source/apex-common/main/classes/fflib_Criteria.cls
```diff
@@ -0,0 +1,654 @@
...
+		implements Evaluator
+{
+	// Add brace when generated to SQL
+	private Boolean embraced = false;
+
+	// List to hold all evaluators
+	private List<Evaluator> evaluators;
+
+	// The type of comparator NONE | AND | OR
+	private String type;
```
SWhalen-gh, line 36: Can this variable name be more descriptive? Consider “private String conjunction”.

sfdx-source/apex-common/main/classes/fflib_Criteria.cls
```diff
@@ -0,0 +1,654 @@
...
+	 * populate private variables with default settings.
+	 */
+	public fflib_Criteria()
+	{
+		this.evaluators = new List<Evaluator>();
+		this.type = 'AND';
+	}
+
+	/**
+	 * Changes the default comparator for each criteria to OR
```
SWhalen-gh, line 49: Is it more accurate to say “changes the default comparator for *the next* criteria...”?

wimvelzeboer: @SWhalen-gh It will change the default comparator for the entire criteria not only the next. Or you need to combine an AND and OR in one query it should be encapsulated between braces. e.g.:
`something AND (x OR y)`. The `x OR y` is another instance of fflib_Criteria, added via the method `addOrCriteria`

sfdx-source/apex-common/main/classes/fflib_Criteria.cls
```diff
@@ -0,0 +1,654 @@
...
+			result += evaluator.toSOQL();
+		}
+
+		if (this.embraced) result = '(' + result + ')';
+
+		return result;
+	}
+
+	private static String operatorToString(fflib_Operator operator)
+	{
```
SWhalen-gh, line 440: Can this be a method (or map)  in the new fflib_Operator class?

sfdx-source/apex-common/main/classes/fflib_Criteria.cls
```diff
@@ -0,0 +1,654 @@
...
+	 * @return Returns the given value converted to literal string
+	 */
+	private static String toLiteral(final Object value)
+	{
+		if (value == null) return 'null';
+
+		if (value instanceof String || value instanceof Id)
+		{
+			String manipulated = (String) value;
+			return '\'' + manipulated + '\'';
```
SWhalen-gh, line 489: Can the class have a toggle for adding “String.EscapeSingleQuotes()” for the string data type? For example,
If (SOQLInjectionProtection == true) manipulated = String.EscapeSingleQuotes(manipulated);

sfdx-source/apex-common/main/classes/fflib_Criteria.cls
SWhalen-gh: That explanation clarifies things.  But these blocks of code are dangerously similar:

.equalTo(...)
.equalTo(...)
.orCriteria(...)   
.equalTo();

//compared to:
.equalTo(...)
.equalTo(...)
.addOrCriteria(...)
.equalTo(...)

It would be clearer if the names of the methods were more distinct, like “setDefaultComparator()” instead of “orCriteria()” ...Or, have “orCriteria()” throw an error if it is not the first statement in the block.

wimvelzeboer: It would look like:
```
new fflib_Criteria()
    .equalTo(...)
    .equalTo(...)
    .addOrCriteria(
         new fflib_Criteria()
             .equalTo(..)
             .equalTo(..)
   )
```
Would result in something like: x AND y AND  (k OR m)

```
new fflib_Criteria()
    .orCriteria()
    .equalTo(...)
    .equalTo(...)
    .addAndCriteria(
         new fflib_Criteria()
             .equalTo(..)
             .equalTo(..)
   )
```
Would result in something like: x OR y OR (k AND m)

wimvelzeboer: It is also not the intention to have very large and complex conditions. As you can combine the conditions as shown in the example.

It is however a good idea to make sure the orCriteria method should be called first in the statement block.

sfdx-source/apex-common/main/classes/fflib_Criteria.cls
wimvelzeboer: The fflib_Operator class is an Enum. 
In the original version (fflib2.0) there is a fflib_OperatorUtils class, with just only this method, not sure if it is that reuable to justify its own public class.

sfdx-source/apex-common/main/classes/fflib_Criteria.cls
wimvelzeboer: I understand what you mean, but I do not think that belongs to the purpose of this class. I am a huge fan of the SOLID design principles. The purpose of this class is to control/hold a criteria, not to go too deep in all the posibilites of a SOQL statement. Maybe I should try to see if I can split this behaviour into a separate class, as I even think that generating the criteria to a SOQL is already pushing the "limits" of the scope of this class.

sfdx-source/apex-common/main/classes/fflib_Criteria.cls
SWhalen-gh: It is a minor point, but it feels like the Operator class should control its string representations. Alternatively, you could say that various client/consumers should manage the rendering.

sfdx-source/apex-common/main/classes/fflib_Criteria.cls
SWhalen-gh: A separate implementation class might be the way to go, although what is here is so very close to the need.  I think offering SOQL criteria creation without injection protection is risky, especially where the rest of the library delivers CRUD and FLS protection (related to DML and SELECT statements). Not having injection protection could discourage adoption, or encourage local customization that makes upgrading more difficult in the future. Maybe the amount of change can be minimized by putting the specifics of SOQL (and its injection management) in a separate “Evaluator” implementation.

wimvelzeboer: My main concern is that this isn't a "SOQL criteria creation" feature in the first place, it is a generic criteria feature and one of the places where it can be used is for rendering SOQL condition strings. 

The main benefit of this feature it to make complex `IF` conditions more readable:
```
    public Accounts getByLargeAccountsInCountry(String countryName)
    {
        return new AccountsImp(
                getRecords(
                       new AccountCriteria()
                               .ShippingCountryEquals(countryName)
                               .NumberOfEmployeesGreaterThan(numberOfEmployees)
                               . ......
                               . ....
                               . ...
                )
        );
    }
```
In addition to this, as I saw that many of these conditions in Domain and Selectors are the same, you can also generate it to SOQL.
I will look into a separate SOQL generator class, something like: 
`String queryWhereString = fflib_CriteriaSOQL.toSOQL(criteria);`

JAertgeerts (dismissed):

sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls
```diff
@@ -298,6 +298,23 @@ public virtual with sharing class fflib_SObjectDomain
     {
     	return Records;
     }
+
+	/**
+	 * @param criteria The condition the returned record must meet
+	 *
+	 * @return Return the SObject records contained in the domain
+	 */
+	public virtual List<SObject> getRecords(fflib_Criteria criteria)
```
wimvelzeboer, line 307: If the domain layer and trigger handler PR is merged, then I will also move this into the fflib_SObjects class

stohn777: @wimvelzeboer   To which PR are you referring here?

However I think the one you're referring to was merged, and since this PR has conflicts, please take a look at those.

wimvelzeboer: Thanks @stohn777 completely forgot about this one, that I still had to rebase it after the merge of the new domain structure.

I think all looks good now. :-)

stohn777:
@wimvelzeboer  
We discussed the PR during our monthly meeting, and although we like the direction, we're going to hold this PR because we're planning a conversation with @capeterson this month where we hope to kick-off a redesign of the query builder feature.
cc. @daveespo @ImJohnMDaniel

wimvelzeboer:
@stohn777 Ok, thanks for the heads-up! Looking forward to seeing the redesign of the query builder. :-)

JAertgeerts:
> @wimvelzeboer We discussed the PR during our monthly meeting, and although we like the direction, we're going to hold this PR because we're planning a conversation with @capeterson this month where we hope to kick-off a redesign of the query builder feature. cc. @daveespo @ImJohnMDaniel

Did this happen at all? Any output to share? 

As a general point of feedback to the team: It'd be nice to see some more roadmaps or thinking on where to take a few of these features. It feels, to me at least, a lot like the current situation of this repository is open source, but closed behind the scenes decision making on features - with very limited ability to influence it by the community. The promises of "we're working on something" and "we have something planned to talk about the direction we want to take" with then no response for months is frustrating.

daveespo:
See #483

Issue #315, opened 2021-01-14
fflib_SecurityUtilsTest failures in Spring '21 Scratch Orgs


tfuda:
I created a Spring '21 "Preview" scratch org today and ran our application's Apex unit tests. Our apps include fflib-apex-common. I observed three failures in fflib_SecurityUtilsTest of the following nature:

```
Apex log for fflib_SecurityUtilsTest.readonly_field_access
==========================================================
47.0 APEX_CODE,DEBUG;APEX_PROFILING,INFO;CALLOUT,INFO;DB,INFO;NBA,INFO;SYSTEM,DEBUG;VALIDATION,INFO;VISUALFORCE,INFO;WAVE,INFO;WORKFLOW,INFO
02:58:51.2 (2784413)|EXECUTION_STARTED
02:58:51.2 (2810336)|CODE_UNIT_STARTED|[EXTERNAL]|01pR0000009SXYy|fflib_SecurityUtilsTest.readonly_field_access()
02:58:51.2 (11760463)|SOQL_EXECUTE_BEGIN|[38]|Aggregations:0|SELECT id, Name FROM Profile WHERE Name = :tmpVar1
02:58:51.2 (16871071)|SOQL_EXECUTE_END|[38]|Rows:0
02:58:51.2 (17496051)|FATAL_ERROR|System.QueryException: List has no rows for assignment to SObject

Class.fflib_SecurityUtilsTest.setupTestUser: line 38, column 1
Class.fflib_SecurityUtilsTest.readonly_field_access: line 56, column 1
02:58:51.2 (17526956)|FATAL_ERROR|System.QueryException: List has no rows for assignment to SObject

Class.fflib_SecurityUtilsTest.setupTestUser: line 38, column 1
Class.fflib_SecurityUtilsTest.readonly_field_access: line 56, column 1
```

It turns out that Salesforce decided that, starting in Spring '21, they will no longer include the "Read Only" Profile in new orgs: [https://help.salesforce.com/articleView?id=release-notes.rn_profiles_and_perms_read_only_new.htm&release=230&type=5](url). The new Profile they are providing is called "Minimum Access - Salesforce", but, unfortunately it cannot be used as a direct substitute for "Read Only", because it doesn't grant "Read" permission to any standard objects. 

[...]

PR #316, opened 2021-01-14
#315 Address the unit test failures resulting from the removal of the "Read Only" Profile

tfuda:
The summary here is that I create an ad-hoc Permission Set that grants read access to Account, Contact and Lead in the testSetup method. Then, I modified the existing setupTestUser method so that if the "Read Only" Profile is absent from the org, it assumes we're in a Spring '21 org, and it uses the "Minimum Access - Salesforce" Profile instead of "Read Only". It then assigns the ad-hoc Permission Set to the User, because "Minimum Access - Salesforce", doesn't grant read access to any standard objects. I also had to create FieldPermissions to grant read access to a couple of fields on Contact in order to satisfy the FLS checks in the readonly_objectAndField_access test method.

### DIFF
--- a/sfdx-source/apex-common/test/classes/fflib_SecurityUtilsTest.cls
+++ b/sfdx-source/apex-common/test/classes/fflib_SecurityUtilsTest.cls
@@ -26,7 +26,76 @@
 
 @isTest
 private class fflib_SecurityUtilsTest {
+
+	@TestSetup
+	static void testSetup() {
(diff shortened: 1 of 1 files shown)


tfuda:
@stohn777 @daveespo If anyone has any thoughts for a better way to do this, I'm all ears.

daveespo (approved):

ImJohnMDaniel (approved):

ImJohnMDaniel:
@tfuda -- Thanks for the fix!

PR #356, opened 2021-08-12
Changes to getChangedRecords

dnyaneshkc:
Added condition for making the method usable in both insert and update contexts. This will reduce the subscribers' code as the framework will check the conditions.

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/356)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls
@@ -289,6 +289,10 @@ public virtual with sharing class fflib_SObjectDomain
 	 **/
 	public List<SObject> getChangedRecords(Set<String> fieldNames)
 	{
+		if(Trigger.isInsert) { // using trigger context variable because I couldn't find an alternative
+			return Records;
+		}
+
 		List<SObject> changedRecords = new List<SObject>();
 		for (SObject newRecord : Records)
 		{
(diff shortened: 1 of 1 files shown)


wimvelzeboer:
Hi @dnyaneshkc 
Could you give an example of when this would be useful? Because I do not really understand this change, as there are no changed records at an insert trigger event.
I am getting the feeling that you might be using the framework incorrectly, having a few example could really help.

dnyaneshkc:
> Hi @dnyaneshkc
> Could you give an example of when this would be useful? Because I do not really understand this change, as there are no changed records at an insert trigger event.
> I am getting the feeling that you might be using the framework incorrectly, having a few example could really help.

Usually we have a piece of code where we're performing the same logic on both insert and update events; validating values, keeping any parent/related records in sync, setting default values (for example: we're setting the Opportunity name in a specific format which needs to be enforced on creation as well as update). These are just a few of the scenarios among many which we have in our project.

We like to re-use the code as much as possible and having the if/else blocks in the subscribers' code simply makes it more difficult to understand for future developers. As a framework provider, I understand where you're coming from but I think it would be a nice to have feature. If the framework took care of both the insert/update scenarios, the subscribers' code will be cleaner and easier to maintain.

Do let me know your thoughts.

Cheers!

wimvelzeboer:
@dnyaneshkc 
Ohk, I understand. 
How about if you would to apply the following pattern?
```
public override void onBeforeInsert()
{
  doSomethingNice()  
}

public override void onBeforeUpdate(...)
{
  onChangedStatusDoSomethingNice()
}

private void onChangedStatusDoSomethingNice()
{
    List<Case> changedRecords = getChangedRecords(new Set<SObjectField>{ Case.Status });
    
    if (changedRecords.isEmpty()) return;
    
    new Cases(changedRecords).doSomethingNice();
}

private void doSomethingNice()
{
  System.debug('Hello There');
}
```
Then you split the logic that is specific to the trigger change event from what you are actually wanting to do. And you re-use the common part without having to duplicate it.

dnyaneshkc:
@wimvelzeboer yeah I think that would help. Still a lot of code on the subscribers' end IMPO but should work. Thank you!

PR #364, opened 2021-09-20
Convert key search to lower case

chivalry:
As far as I can tell, the check for the key being in the instance cache will never be executed. You're creating `fflib_SObjectDescribe` instances in three places:

- `fflib_SObjectDescribe.getDescribe(String)` (line 193)
- `fflib_SObjectDescribe.getDescribe(Schema.SObjectType)` (line 202)
- `fflib_SObjectDescribe.getDescribe(Schema.DescribeSObjectResult)` (line 210)

Each time you're only doing so after you've checked if the token (lower cased) is in the instance cache and only create the new instance if it's not.

But maybe you're directly instantiating this class in other places. If not, the check (and the exception) can be removed. If so, then this check should have the token converted to lower case.

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/364)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_SObjectDescribe.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjectDescribe.cls
@@ -77,7 +77,7 @@ public class fflib_SObjectDescribe {
 	private fflib_SObjectDescribe(Schema.SObjectType token){	
 		if(token == null)
 			throw new InvalidDescribeException('Invalid SObject type: null');
-		if(instanceCache.containsKey( String.valueOf(token) ))
+		if(instanceCache.containsKey( String.valueOf(token).toLowerCase() ))
 			throw new DuplicateDescribeException(token + ' is already in the describe cache');
 		this.token = token;
 		instanceCache.put( String.valueOf(token).toLowerCase() , this);
--- a/sfdx-source/apex-common/main/classes/fflib_SObjectDescribe.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjectDescribe.cls
@@ -77,8 +77,6 @@ public class fflib_SObjectDescribe {
 	private fflib_SObjectDescribe(Schema.SObjectType token){	
 		if(token == null)
 			throw new InvalidDescribeException('Invalid SObject type: null');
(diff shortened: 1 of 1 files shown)


daveespo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/364)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @chivalry and @daveespo)

*[a discussion](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/364#-Mk7R3p4-lbJEyhXfpJN:-Mk7R3p5B7gcfG6RmMjF:b688gif) (no related file):*
@chivalry -- can you provide a repro case (Anonymous Apex would be fine) to demonstrate the bug you're fixing with this PR?

---

chivalry:
I'll look into doing that, but it's not actually simple to do so. My actual point about this is that the check, as it currently is written, doesn't have a code path that will execute line 81, and it might be better to simply remove lines 80 and 81. If I'm incorrect about that, then maybe it makes sense to keep the check for throwing the error, but this check makes sure the key isn't already in the `instanceMap`, but each time you create a new instance of `fflib_SObjectDescribe`, you've already checked that there isn't.

wimvelzeboer:
Hi @chivalry 
The constructor is private and therefore cannot be called from any other place. There are currently just the three places you mentioned. All three are doing a toLowerCase operation first prior to checking if the value is already in the `instanceCache`.
The keys of this map are all in lower case.

So, I agree with you that line 80 doesn't really make sense. 
In the current code, it will never find a duplicate, and never hit that exception. That exception is indeed dead-code.

And I agree with you, we either should add the toLowerCase in anticipation of future changes to this class or remove the IF condition and exception entirely.

@daveespo This doesn't seem to be a bug nor can we hit this piece of code with anonymous apex, its more a matter of cleaning up or removing dead-code.

My preference would be to remove the lines 80-81.

chivalry:
> My preference would be to remove the lines 80-81.

Me too, actually, but that may be because part of my job is primary code reviewer on our project, and I'm kind of nitpicky about things like that.

PR #365, opened 2021-09-23
Created fflib_WhereBuilder as a helper for fflib_QueryFactory

jonnypetraglia:
A simple class to avoid (part of) String concatenation. `setCondition`.

  1. Greater confidence with the compiler - Just like QueryFactory, this is compiled, which means even less of a chance for joining strings improperly.
  2. Easier testing - Instead of "expect the query to be this EXACT thing" (when order of things doesn't always matter) whereas now we can say "Assert that the WhereBuilder contains this exact set of criteria.
  3. Nested criteria - Huge benefit in my opinion to be have the _compiter_ match paren.
  4. Presumably, performance - Granted, you'd get similar with a fflib_StringBuilder, but still, it's worth noting that avoiding raw string concatenation is groovy.
  5. Readability - Check the examples below.

```
fflib_QueryFactory queryFactory = newQueryFactory(true);
// Before
queryFactory.setCondition('Id in :subjectIds AND (AssignedSalesRep__c == null OR AssignedSalesRep__r.Name = \'Rusty Shackleford\')');
// After
queryFactory.setCondition(fflib_WhereBuilder
    .whereX('Id in :subjectIds')
    .andX(fflib_WhereBuilder
        .orX(
             'AssignedSalesRep__c == null',
             'AssignedSalesRep__r.Name = \'Rusty Shackleford\''
        )
    ).toString());
// Or inline
queryFactory.setCondition(fflib_WhereBuilder.whereX('Id in :subjectIds').andX(fflib_WhereBuilder.orX('AssignedSalesRep__c == null', 'AssignedSalesRep__r.Name = \'Rusty Shackleford\'')).toString());
// Or extract-refactor out query building into function(s) as you wish
```

If this is approved, I'd be willing to add an overload to QueryFactory for `setCondition(fflib_WhereBuilder);` before the merge.

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_IWhereBuilder.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_IWhereBuilder.cls
@@ -1,0 +1,6 @@
+public interface fflib_IWhereBuilder {
+    fflib_IWhereBuilder andX(String condition);
+    fflib_IWhereBuilder andX(fflib_IWhereBuilder condition);
+    fflib_IWhereBuilder orX(String condition);
+    fflib_IWhereBuilder orX(fflib_IWhereBuilder condition);
+}
--- a/sfdx-source/apex-common/main/classes/fflib_IWhereBuilder.cls-meta.xml
+++ b/sfdx-source/apex-common/main/classes/fflib_IWhereBuilder.cls-meta.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ApexClass xmlns="http://soap.sforce.com/2006/04/metadata">
+    <apiVersion>52.0</apiVersion>
+    <status>Active</status>
+</ApexClass>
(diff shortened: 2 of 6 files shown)


wimvelzeboer:
Hi @jonnypetraglia 
This might also be a great alternative: 
https://github.com/wimvelzeboer/fflib-apex-extensions/wiki/criteria-based-filter
That solution is now part of a extensions package, there is also a PR #313 open with the same.

With this you can reuse the same conditions for selectors and filters for domains.

ImJohnMDaniel:
@jonnypetraglia and @wimvelzeboer -- the idea of including a "where clause builder" framework has been approached a couple of different times in the FFLIB-Apex-Common project over the years.  Both of your approaches are interesting and address the fundamental idea.  We have seen previous PRs and similar efforts separate from the project. 

I wanted to reach out and let you know that we have been discussing the idea quite a bit over the last two months. When we are ready, we want to solicit feedback on its construction from the community as from other folks, including the `fflib_QueryFactory`'s original creator @capeterson.

For the moment, we are putting these types of PRs on hold until we can determine the best path forward.  Thanks for your patience.

cc: @stohn777, @daveespo, @afawcett

PR #387, opened 2021-11-17
Fix bug in the getChangedRecords method

wimvelzeboer:
A colleague at work found an issue with the getChangedRecords method, it is not listing records as changed when only a casing is different in a field. e.g. LastName field changed from "smith" to "Smith".

e.g.:
This fails: as it outputs "Not Changed"'
```java
Contact r1 = new Contact(LastName = 'Test');
Contact r2 = new Contact(LastName = 'test');
SObjectField fieldName = Contact.LastName;
if (r1.get(fieldName) != r2.get(fieldName))
{
    System.debug('Changed');
} else 
{
   System.debug('Not Changed');
}
```
while this works:
```java
Contact r1 = new Contact(LastName = 'Test');
Contact r2 = new Contact(LastName = 'test');
SObjectField fieldName = Contact.LastName;
if (r1.get(fieldName) !== r2.get(fieldName))
{
    System.debug('Changed');
} else 
{
   System.debug('Not Changed');
}
```

So, instead of `!=` we need to use `!==`

CC: @stohn777 @daveespo @ImJohnMDaniel

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/387)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls
@@ -301,7 +301,10 @@ public virtual with sharing class fflib_SObjectDomain
 			SObject oldRecord = this.ExistingRecords.get(recordId);
 			for (String fieldName : fieldNames)
 			{
-				if (oldRecord.get(fieldName) != newRecord.get(fieldName))
+				Object oldValue = oldRecord.get(fieldName);
+				Object newValue = newRecord.get(fieldName);
+				if (((oldValue instanceof String || newValue instanceof String)	&& oldValue !== newValue)
+						||	oldValue != newValue)
 				{
 					changedRecords.add(newRecord);
 					break;  // prevents the records from being added multiple times
@@ -328,7 +331,10 @@ public virtual with sharing class fflib_SObjectDomain
 			SObject oldRecord = this.ExistingRecords.get(recordId);
 			for (Schema.SObjectField fieldToken : fieldTokens)
 			{
-				if (oldRecord.get(fieldToken) != newRecord.get(fieldToken))
+				Object oldValue = oldRecord.get(fieldToken);
+				Object newValue = newRecord.get(fieldToken);
+				if (((oldValue instanceof String || newValue instanceof String)	&& oldValue !== newValue)
+						||	oldValue != newValue)
 				{
 					changedRecords.add(newRecord);
 					break;  // prevents the records from being added multiple times


daveespo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/387)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @daveespo and @wimvelzeboer)

___
*[sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls, line 304 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/387#-Mohpbg2ANezu9Jt5JCG:-Mohpbg2ANezu9Jt5JCH:bbkkyzd) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/2009370a4ccbde89d38a90d88bf899b5be6474dc/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls#L304)):*
> ```VBA
> 			for (String fieldName : fieldNames)
> 			{
> 				if (oldRecord.get(fieldName) !== newRecord.get(fieldName))
> ```

The triple-equals ensure that the two values are the exact same instance, not that they are equivalent. 

While that might work for String, it doesn't work for other Objects. 

```
Opportunity o1 = new Opportunity(CloseDate=Date.newInstance(2022,1,1));
Opportunity o2 = new Opportunity(CloseDate=Date.newInstance(2022,1,1));
System.assert(o1.get('CloseDate') == o2.get('CloseDate')); //Pass
System.assert(o1.get('CloseDate') === o2.get('CloseDate')); //Fail
```

This solution needs reworking

wimvelzeboer:
> While that might work for String, it doesn't work for other Objects.
> 
> ```
> Opportunity o1 = new Opportunity(CloseDate=Date.newInstance(2022,1,1));
> Opportunity o2 = new Opportunity(CloseDate=Date.newInstance(2022,1,1));
> System.assert(o1.get('CloseDate') == o2.get('CloseDate')); //Pass
> System.assert(o1.get('CloseDate') === o2.get('CloseDate')); //Fail
> ```
> 
> This solution needs reworking
Ah, I should have thought of that.. 
I now pushed another change that does fix it for string fields.

ImJohnMDaniel (approved):

stohn777 (approved):
I'm good with this insightful workaround, however I'm curious whether all of the field types implement an `.equals(...)` method that we could leverage because I know that `String.equals(...)` would detect the difference in case.

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/387#-:-Moj7o3e1L1pidtXnHvN:b-ql935t)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @daveespo and @wimvelzeboer)

stohn777:
Perhaps not.  Evidently `Object.equals(...)` with a String does not work as this logic would need.  Explicit code for a String would still be needed.
```
Object base = 'true';
Object same = 'true';
Object diff = 'True';

// Passes
System.assert(base.equals(same), 'Values are different.  Expected: ' + base.toString() + ' Actual: ' + same.toString());

// Fails
System.assert(!base.equals(diff), 'Values are equal.  Expected: ' + base.toString() + ' Actual: ' + diff.toString());
```

stohn777 (commented):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/387)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ImJohnMDaniel and @wimvelzeboer)

___
*[sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls, line 304 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/387#-Mohpbg2ANezu9Jt5JCG:-Moo7T88DQSoh2xM5VUT:bwlxazl) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/2009370a4ccbde89d38a90d88bf899b5be6474dc/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls#L304)):*

A few things with this reworked solution:

 * We need test coverage for this scenario. The test needs to cover the String cases as well as other instance types like Date, and numerics (`Opportunity` has all of these field types so I think it'll be a good candidate for thorough test coverage) 
 * This code change is working around an extremely subtle bug with Apex. Assigning a String to an Object causes the documented behavior of String.equals() to change depending on the left side of the assignment. It's "documented" over here https://salesforce.stackexchange.com/questions/339160/why-doesnt-equals-behave-like-a-virtual-method and https://github.com/ChuckJonas/wtfapex#object-equals-override so please add a code comment above this changed section to reflect reason for this unconventional conditional
 * We've definitely crossed the line of DRY here .. we now have two methods that are identical in every way except the method signature; given the nature of this bug and what I'm asking for above, can you please isolate the body of this for-loop into a private utility method that is reused by both the `String fieldName` and `SObjectField fieldToken` overload of this method?
 * I have some concern about regressive impact of this change -- given how inconsistently string comparisons are implemented in Apex, are there users out there that will be surprised or negatively impacted by this  change? @stohn777  and @imjohnmdaniel  -- thoughts?


 

</blockquote></details>

@daveespo, really good points.  Expounding thereupon.

**Bullet 3**
This could easily be reduced to a single method of implementation in addition to a simple overload.

**Bullet 4**
Regressive impact is an insightful concern.  Perhaps an injection based property could be introduced for the String comparison logic the org wants with the default being the older broken functionality, although I'm open to debate for fixing Apex's shortcoming and requiring those that want the broken logic to do the little bit of extra work.  With this, maintain backward compatibility is available while accommodating those wanting a more canonical string comparison.

**Thought-Provoking Code**
```
interface fflib_IStringComparison
{
    Boolean equals(String value1, String value2);
}
```
```
public virtual with sharing class fflib_SObjectDomain

...

public fflib_IStringComparison stringCompareImplementation = new fflib_SObjectDomain.StringComparison();

public class StringComparison
    implements fflib_IStringComparison
{
    public Boolean equals(String value1, String value2)
    {
        return value1 == value2;
    }
}

...

if (((oldValue instanceof String || newValue instanceof String) && !stringCompareImplementation.equals(oldValue, newValue))

...
```

daveespo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/387)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ImJohnMDaniel, @stohn777, and @wimvelzeboer)

___
*[sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls, line 304 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/387#-Mohpbg2ANezu9Jt5JCG:-MqGqvT18j0qvSdChOGs:bsmfci6) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/2009370a4ccbde89d38a90d88bf899b5be6474dc/sfdx-source/apex-common/main/classes/fflib_SObjectDomain.cls#L304)):*

@daveespo, really good points.  Expounding thereupon.

**Bullet 3**
This could easily be reduced to a single method of implementation in addition to a simple overload.

**Bullet 4**
Regressive impact is an insightful concern.  Perhaps an injection based property could be introduced for the String comparison logic the org wants with the default being the older broken functionality, although I'm open to debate for fixing Apex's shortcoming and requiring those that want the broken logic to do the little bit of extra work.  With this, maintain backward compatibility is available while accommodating those wanting a more canonical string comparison.

**Thought-Provoking Code**
```
interface fflib_IStringComparison
{
    Boolean equals(String value1, String value2);
}
```
```
public virtual with sharing class fflib_SObjectDomain

...

public fflib_IStringComparison stringCompareImplementation = new fflib_SObjectDomain.StringComparison();

public class StringComparison
    implements fflib_IStringComparison
{
    public Boolean equals(String value1, String value2)
    {
        return value1 == value2;
    }
}

...

if (((oldValue instanceof String || newValue instanceof String) && !stringCompareImplementation.equals(oldValue, newValue))

...
```
</blockquote></details>

Hi @wimvelzeboer  -- I'm sorry we took so long to respond on this thread -- December is always a tough month  to get anything done! 

We discussed this bug report at length and have decided that the shape of this PR should address the concern raised by my prior comment about accommodating for users who have come to depend on the existing behavior of case-insensitive String matching (i.e. we can't introduce breaking changes without providing a workaround). So, concretely, we'd like to see:

 * Define a new `virtual protected` method on `fflib_SObjectDomain` or `fflib_SObjects` named something like `protected virtual Boolean objectEquals(Object left, Object right)`
 * The default implementation of `objectEquals` should implement the logic you have in this PR (i.e. it should use the `==` to compare all types except String where you should use `===` (triple equals). 
 * Add an Apexdoc style comment above this method explaining that it's intended to be overridden by consumers that expect case-insensitive String comparison and link to this Github issue (PR) along with  https://salesforce.stackexchange.com/questions/339160/why-doesnt-equals-behave-like-a-virtual-method and https://github.com/ChuckJonas/wtfapex#object-equals-override explaining why we don't just use `.equals()`
 * We need test coverage of the new method and of getChangedRecords that covers the new case-sensitive expectation of string comparisons
 * Most of the concerns I had about DRYing up the duplicated code will go away when you implement the new method but if you can stretch a bit further to have the overloaded version of `getChangedRecords(Set<FieldToken> tokens)` delegate to the `getChangedRecords(Set<String> fieldNames)`, that would be ideal

Thank you for you contributions as always!

PR #390, opened 2021-12-14
Allow setFieldValueByMap to consume a Map<Id, Id>

JAertgeerts:


ImJohnMDaniel:
@JAertgeerts -- thanks for the PR.  A couple of questions:
* Why is this change needed?
* Can you provide corresponding code coverage please?

wimvelzeboer:
@ImJohnMDaniel The purpose of the domain is to allow easy access to the list of objects it contains. There are three primary types of methods on the domain; getters, setters and filter methods.
The more standard setter methods we have in the fflib_SObjects domain class, the smaller the methods can become on the domain classes (e.g. Accounts).

This method allows for easy setting and Id field based on the provided map. An example would be the following method, which sets the Case.OwnerId based on the provided map. It check if the Case.AccountId is populated in the map, if so, then it takes the value of that key and sets it to the provided target field, the Case.OwernId.
```
public ICases setOwnerIdByAccountId(Map<Id, Id> UserIdByAccountId)
{
   setFieldValueByMap(Case.AccountId, Case.OwnerId, UserIdByAccountId);
   return this;
}
```

Since Salesforce is a bit limited with casting a Map<Object, Object> to a Map<Id, Id> we need another method overload.

wimvelzeboer (approved):

ImJohnMDaniel:
@JAertgeerts, just following up on the request to provide matching code coverage for this method.  Cheers!

daveespo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/390)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ImJohnMDaniel, @JAertgeerts, and @stohn777)

___
*[sfdx-source/apex-common/main/classes/fflib_SObjects.cls, line 294 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/390#-MsWaLEf4PK7M8kMkFPu:-MsWaLEf4PK7M8kMkFPv:b-jfi38a) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/990827963b23fc09f4c233a65861a2ac3fd02079/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L294)):*
> ```VBA
> 	 * @param values Map of values to store by the sObjectFieldToCheck fields value
> 	 */
> 	protected virtual void setFieldValueByMap(
> ```

Agreed with JohnD, we need a test method to cover this new code path

PR #409, opened 2022-05-10
Add method to include ALL ROWS in QueryFactory

1903Daniel:
We have a use case in our project using FFLIB where we need to query for archived and deleted records as well.

The proposed change adds the ability to call `setAllRows(true)` on QueryFactory objects, which results in ALL ROWS being appended to the resulting query. It is also possible to reset the property with `setAllRows(false)` after e.g. deep cloning a QueryFactory.

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/409)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
@@ -666,8 +666,8 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 	/**
 	 * @param addAllRows whether an ALL ROWS clause will be added to the resulting query.
 	**/
-	public fflib_QueryFactory setAllRows(Boolean addAllRows){
-		this.allRows = addAllRows;
+	public fflib_QueryFactory setAllRows(){
+		this.allRows = true;
 		return this;
 	}
 
@@ -713,8 +713,9 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 		if(offsetCount != null)
 			result += ' OFFSET '+offsetCount;
 		
-		if(allRows)
+		if(allRows) {
 			result += ' ALL ROWS';
+		}
 
 		return result;
 	}
@@ -728,7 +729,6 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 		fflib_QueryFactory clone = new fflib_QueryFactory(this.table)
 			.setLimit(this.limitCount)
 			.setOffset(this.offsetCount)
-			.setAllRows(this.allRows)
 			.setCondition(this.conditionExpression)
 			.setEnforceFLS(this.enforceFLS);
 
--- a/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
@@ -664,7 +664,7 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 	}
 
 	/**
-	 * @param addAllRows whether an ALL ROWS clause will be added to the resulting query.
+	 * whether an ALL ROWS clause will be added to the resulting query
 	**/
 	public fflib_QueryFactory setAllRows(){
 		this.allRows = true;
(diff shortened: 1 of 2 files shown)


1903Daniel:
@ImJohnMDaniel Thanks for the comment. As per documentation, ALL ROWS doesn't control which fields should be retrieved. Rather it controls which **records** should be retrieved (deleted and archived records as well): https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/langCon_apex_SOQL_query_all_rows.htm

So the responsibility for specifying the fields still lies at the developer implementing the Selector class.

1903Daniel:
@ImJohnMDaniel Thanks a lot 😃 I appreciate the review process

sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
```diff
@@ -662,6 +663,14 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 		return setOrdering(ordr);
 	}
 
+	/**
+	 * @param addAllRows whether an ALL ROWS clause will be added to the resulting query.
+	**/
+	public fflib_QueryFactory setAllRows(Boolean addAllRows){
```
ImJohnMDaniel, line 669: @1903Daniel -- I would suggest just a no parameter method.  Simply `fflib_QueryFactory setAllRows()` 

In this case, you would only ever flip the `allRows` class variable to `true`.  Since it defaults to `false`, there is no real benefit to calling this method with a parameter of `false`.  

I would suggest changing this method to no parameters and then simply set the `allRows` member variable to true.

ImJohnMDaniel (changes requested):

sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
1903Daniel: @ImJohnMDaniel Thanks for the suggestion. I thought about this as well. However, after you `deepClone()` an `fflib_QueryFactory` with the variable already set to `true`, you wouldn't be able to flip the variable back to `false` again. Do you think this is a valid use case? If not, I'll happily change `setAllRows` to a no params method.

daveespo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/409)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @1903Daniel, @ImJohnMDaniel, and @stohn777)

___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 716 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/409#-N2kyrYn4KAC5SfK1bCX:-N2kyrYn4KAC5SfK1bCY:b-3yrfhe) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/0016af10c89e207d36e57a8af5238e2bdcdc6311/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L716)):*
> ```VBA
> 			result += ' OFFSET '+offsetCount;
> 		
> 		if(allRows)
> ```

Please use curly braces even for 1 line conditionals (and I'll acknowledge that this project is littered with violations of that rule .. but let's not keep digging the hole deeper)

sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
ImJohnMDaniel: @1903Daniel, I cannot say that I have ever done a `deepClone()` of a QueryFactory instance.  Let's proceed with the change to a no param method.  If someone comes across this issue in the future, then we can deal with it then.  Cheers!

1903Daniel: @ImJohnMDaniel Done

daveespo (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/409)* status: all files reviewed, 1 unresolved discussion (waiting on @ImJohnMDaniel and @stohn777)

ImJohnMDaniel (approved):

ImJohnMDaniel:
@1903Daniel -- Thanks for the help!!

PR #413, opened 2022-05-27
Make sure method return non-null values

wimvelzeboer:
After upgrading my projects to the latest fflib-apex-common, where some methods were moved from fflib-apex-extensions to fflib-apex-common, I found that a high number of unit-tests were failing.

Then I found out that the `getIdFieldValues` and `getStringFieldValues` were suddenly also returning null values, which makes no sense. This fix will make sure these methods always return the values and omit the nulls


sfdx-source/apex-common/main/classes/fflib_SObjects.cls
```diff
@@ -161,11 +161,13 @@ public virtual class fflib_SObjects
 	 * @return Return a set with all the Id values of the given field
 	 */
 	@TestVisible
-	protected Set<Id> getIdFieldValues(Schema.SObjectField field)
+	protected virtual Set<Id> getIdFieldValues(Schema.SObjectField field)
 	{
 		Set<Id> result = new Set<Id>();
 		for (SObject record : getRecords())
 		{
+			if (record.get(field) == null) continue;
```
ImJohnMDaniel, line 169: @wimvelzeboer please add braces for readability.  Otherwise, looks good to me.

sfdx-source/apex-common/main/classes/fflib_SObjects.cls
```diff
@@ -177,11 +179,13 @@ public virtual class fflib_SObjects
 	 * @return Return a set with all the String values of the given field
 	 */
 	@TestVisible
-	protected Set<String> getStringFieldValues(Schema.SObjectField field)
+	protected virtual Set<String> getStringFieldValues(Schema.SObjectField field)
 	{
 		Set<String> result = new Set<String>();
 		for (SObject record : getRecords())
 		{
+			if (record.get(field) == null) continue;
```
ImJohnMDaniel, line 187: @wimvelzeboer please add braces for readability.

ImJohnMDaniel (changes requested):

ImJohnMDaniel:
@wimvelzeboer, can you also please check `fflib_SObjectsTest.itShouldReturnFieldValues` test method.  It failed in the CI run for this PR

sfdx-source/apex-common/main/classes/fflib_SObjects.cls
```diff
@@ -161,11 +161,16 @@ public virtual class fflib_SObjects
 	 * @return Return a set with all the Id values of the given field
 	 */
 	@TestVisible
-	protected Set<Id> getIdFieldValues(Schema.SObjectField field)
+	protected virtual Set<Id> getIdFieldValues(Schema.SObjectField field)
 	{
 		Set<Id> result = new Set<Id>();
 		for (SObject record : getRecords())
 		{
+			if (record.get(field) == null)
```
ImJohnMDaniel, line 169: @wimvelzeboer, would using `Sobject.isSet(Schema.SObjectField)` be more appropriate here?

[ref](https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_methods_system_sobject.htm#apex_System_SObject_isSet_2)

sfdx-source/apex-common/main/classes/fflib_SObjects.cls
wimvelzeboer: @ImJohnMDaniel you are indeed right, I changed it!

sfdx-source/apex-common/main/classes/fflib_SObjects.cls
wimvelzeboer: @ImJohnMDaniel You are right indeed! Its changed!

ImJohnMDaniel:
@wimvelzeboer, the `fflib_SObjectsTest.itShouldReturnFieldValues()` test method is still failing with the following error
```
System.AssertException: Assertion Failed                                
Class.fflib_SObjectsTest.itShouldReturnFieldValues: line 146, column 1  
```

Can you take a look at this please?

daveespo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/413)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @ImJohnMDaniel, @stohn777, and @wimvelzeboer)

*[a discussion](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/413#-N91UxHoBgW--LSgDls7:-N91UxHoBgW--LSgDls8:b-z6574g) (no related file):*
Due to the changes in behavior that this PR proposes, my vote is to reject it

___
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 169 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/413#-N34WjFibFt8vCYX5NnN-r2-169:-N91UP_L8KVWfKLOaHUH:bu9g9xv) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/5dbfb3a4ab87fd64f7e0ca4170ebe95d3bde5379/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L169)):*

@ImJohnMDaniel You are right indeed! Its changed!
</blockquote></details>

I'm less concerned about the breaking change that this introduces since you're asking for Id values and those should be non-null. But the risk isn't zero that it will break someone's assumption of how this method works.
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 188 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/413#-N91Th_Q8lfkHyb0I2Ol:-N91Th_R1Ba4GAlnIIpb:b-1ip0vu) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/d0f4f68412919f0984c260056428f9d1ad55cab6/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L188)):*
> ```VBA
> 		for (SObject record : getRecords())
> 		{
> 			if (String.isBlank((String) record.get(field)))
> ```

I don't know if I agree with this change -- Blank strings and null are distinct values for a String field. The Apex doc comment nor does the method name make it clear that values consisting of whitespace only will be excluded. 

This change is definitely a breaking change and not worth the risk, IMO.

PR #420, opened 2022-07-13
#419 - implements a proposal for supporting native User Mode Database Operations introduced in Summer '22

daveespo:
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)


daveespo (commented):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @ImJohnMDaniel and @stohn777)

___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 91 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tPlj183aajktfz1_Q:-N6tPlj183aajktfz1_R:bz0km6b) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/3cf078803a27a4d9198bb4f09c9efe1fccffc6ba/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L91)):*
> ```VBA
> 	private String getFieldPath(String fieldName, Schema.sObjectType relatedSObjectType){
> 
> 		//Enforcing FLS using the legacy heuristic requires resolving the full field path to its respective
> ```

Here's how my reasoning went down: 

We discussed keeping the "User Mode" and "Enforce Schema Validation" as separate concerns so that's where I started. But then I started writing up the Apex Doc comments on why you'd want to leave Schema Validation ON if you WEREN'T enabling FLS and it sounded silly ("Turn this feature on if you need us to validate that you have syntactical errors in your SOQL"). Why would you want to expend computation expense each time you run a query just because you're sloppy?

So I reasoned that we only need to do this heavy lifting if the legacy form of FLS enforcement was enabled.

This change causes many failing tests .. a few for reasons that are actually silly. But there are a few actual logic changes if you turn schema validation off. Notably, standard objects that have a lookup to a parent usually have an "<parent object>Id" field 

For example, this test fails: 

```
13:35:41.189 (1191507628)|FATAL_ERROR|System.AssertException: Assertion Failed: Expected: SELECT CreatedBy.ManagerId, CreatedBy.Name, FirstName, Id, LastModifiedBy.Email, LastName FROM User, Actual: SELECT CreatedBy.Manager, CreatedBy.Name, FirstName, Id, LastModifiedBy.Email, LastName FROM User

Class.fflib_QueryFactoryTest.deterministic_toSOQL: line 555, column 1
```

We do black magic to transpose the "CreatedBy.Manager" into "CreatedBy.Manager**Id**" 

See https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/master/sfdx-source/apex-common/main/classes/fflib_SObjectDescribe.cls#L110

For that reason, we can't introduce this kind of change as backwards-compatible. 

There are other problems too -- the Set of fields (`private Set<String> fields`) is a case-sensitive Set of strings .. so the Set doesn't implicitly de-dupe someone selecting a field using two different cases (ex "Name" and "name" are different elements in the Set and you end up with a 'duplicate field selected' SOQL error)

So now I'm back at trying to figure out how to accomplish the goal of increasing efficiency of query construction (i.e. avoiding the Describe loops) without creating immense challenges for people adopting this change.
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls` line 117 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tSSBRAx8r-545X1aA:-N6tSSBRAx8r-545X1aB:b-yu700r) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/3cf078803a27a4d9198bb4f09c9efe1fccffc6ba/sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls#L117)):*
> ```VBA
>     }
> 
>     public fflib_SObjectSelector(Boolean includeFieldSetFields, Boolean userMode)
> ```

We talked about not adding yet-another-overload of the constructor and instead using a Config builder pattern. But we have all of these fluent style state mutation methods (`ignoreCRUD()`, `enforceFLS()`) and once I started trying to add a builder object, it got so messy with trying to retain backwards compat given that "there is more than one way to do it" 
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls` line 125 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tSvmP4sc2j-kr0CJa:-N6tSvmP4sc2j-kr0CJb:b61lk82) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/3cf078803a27a4d9198bb4f09c9efe1fccffc6ba/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls#L125)):*
> ```VBA
>     }
> 
> 	public virtual class UserModeDML extends SimpleDML{
> ```

And we need to talk about if/how we support this from `fflib_Application`. 

There is already a newInstance method to construct a SOUOW and pass it an IDML implementation. Adding a "userMode" overload of newInstance() felt wrong. If you want a UserMode version of SOUOW, you should just use the existing newInstance() method and pass it an instance of this class.

tfuda (commented):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)* status: all files reviewed, 4 unresolved discussions (waiting on @daveespo, @ImJohnMDaniel, and @stohn777)

___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls` line 119 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N70ee2G-v7A8BMAWG2C:-N70ee2H3vHmpBpRR0Ay:b-l48v1o) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/3cf078803a27a4d9198bb4f09c9efe1fccffc6ba/sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls#L119)):*
> ```VBA
>     public fflib_SObjectSelector(Boolean includeFieldSetFields, Boolean userMode)
>     {
>         this(includeFieldSetFields, false, false, false, true);
> ```

Shouldn't you be passing userMode as the last arg here?

afawcett:
I think the challenge here is this PR is attempting to do to many things - its done an excellent job of exposing the user mode capability - but it also challenged itself with optimizing the field validation and related features. I took a copy of this code and commented out the code in `getFieldPath` and all tests pass just fine. 

		/**if(mFlsEnforcement != FLSEnforcement.LEGACY){
			return fieldName;
		}**/

Personally I would conceded to fight that battle over optimizing for another day and run with these changes without the above code. I'll also be transparent and say that it would greatly help my work on the next edition of the book as well! ;-)

afawcett:
I also think we should add... explicit SYSTEM_MODE, as using NONE will still enforce sharing depending on the sharing clause on the outer class. 

    public enum FLSEnforcement{NONE, LEGACY, USER_MODE, SYSTEM_MODE}

afawcett:
I also found I needed to change where the WITH USER_MODE was emitted during the SOQL string build - per docs it needs to happen after the WHERE clause...

		result += ' FROM ' + (relationship != null ? relationship.getRelationshipName() : table.getDescribe().getName());
		if(conditionExpression != null)
			result += ' WHERE '+conditionExpression;

		if(mFlsEnforcement == FLSEnforcement.USER_MODE){
			result += ' WITH USER_MODE';
		}
		if(mFlsEnforcement == FLSEnforcement.SYSTEM_MODE){
			result += ' WITH SYSTEM_MODE';
		}

tfuda (commented):
> Reviewed 2 of 2 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)* status: all files reviewed, 7 unresolved discussions (waiting on @daveespo, @ImJohnMDaniel, and @stohn777)

___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 301 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJtTn364S6XMSi_Si9:-NFJtTn364S6XMSi_SiA:b-ki7rjr) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/899c3257ac42eb1b4029950568eaa0dfc5082eef/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L301)):*
> ```VBA
> 	**/
> 	public fflib_QueryFactory selectFields(Set<Schema.SObjectField> fields){
> 		for(Schema.SObjectField token:fields){
> ```

Couldn't / shouldn't this just be:
```
public fflib_QueryFactory selectFields(Set<Schema.SObjectField> fields) {
   for (Schema.SObjectField token : fields) {
      selectField(token);
   }
   return this;
}
```
___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 314 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJulHk2Xa44kNyYZZu:-NFJulHk2Xa44kNyYZZv:bmbzn1e) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/899c3257ac42eb1b4029950568eaa0dfc5082eef/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L314)):*
> ```VBA
> 	/**
> 	 * Selects multiple fields. This acts the same as calling {@link #selectField(Schema.SObjectField)} multiple times.
> 	 * @param fields the set of {@link Schema.SObjectField}s to select.
> ```

Should be "the LIST of {@link Schema.SObjectField}s to select."
___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 318 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJv5J-4O4Ax_dyoKC3:-NFJv5J-4O4Ax_dyoKC4:b-gmrncq) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/899c3257ac42eb1b4029950568eaa0dfc5082eef/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L318)):*
> ```VBA
> 	**/
> 	public fflib_QueryFactory selectFields(List<Schema.SObjectField> fields){
> 		for(Schema.SObjectField token : fields){
> ```

Same comment as above; can't this delegate to the selectField(Schema.SObjectField) method to DRY the code a bit?
___
*[`sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls` line 376 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJwGlk0Jfx7rNz20su:-NFJwGlk0Jfx7rNz20sv:blmdp31) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/8c30693581c591a696b7ad3bcf35585fd12bc651/sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls#L376)):*
> ```VBA
> 
> 	@IsTest
> 	static void toSoql_When_UserModeAndUserCannnotReadObject_Expect_QueryException(){
> ```

Minor nit; the method under test is called toSOQL, but your test methods are called toSoql_XXX

stohn777 (approved):
Cloned and pushed to a scratch org for some playing and testing.  Looks good.

afawcett:
Thanks all - all this makes sense - look forward to seeing this merged. I currently have a copy of fflib and mocks in my book sample app repo. @ImJohnMDaniel and I are keen to remove this and simply pull latest from these repos via setup scripts. @ImJohnMDaniel is working on those scripts for me (thanks dude!). Are we able to get this merged in the next month or so max?

daveespo (commented):
I believe this PR is now merge-ready. I added more test cases based on the bugs I found while running this branch against my largest project (1900 test methods) which exposed bugs with how namespaces are handled (our project is a managed package). 

The legacy test methods and coverage need some love -- mostly to clean up naming conventions and make it clear what the method is actually testing -- and at the same time, run some of the existing test methods against "SYSTEM_MODE" versions of the same Selectors to prove out that the assertions that those test methods make are valid in both cases -- but that's too much churn for this PR. 

We can't actually integrate this branch into my production managed package until User Mode in Apex is GA -- presumably with Spring '23 (Feb 2023) -- so I can't justify the expense to QA the app with the new SYSTEM_MODE and USER_MODE variants -- the best I can do in the short term is to confirm that test coverage passes (which it does with a few minor asterisks which are problems on my end with how we're enumerating fields for "SELECT *" scenarios) 

> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-:-NI9C42F1P1P2cPQzwbn:b-9irr1p)* status: 3 of 5 files reviewed, 11 unresolved discussions (waiting on @ImJohnMDaniel and @tfuda)

___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 91 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tPlj183aajktfz1_Q:-NFJtOXN2c2ea9Nv2vGs:bmtu1lb) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/4c180b1c456d8c38dee3b1f5a46266d2386a4ab1/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L91)):*

Here's how my reasoning went down: 

We discussed keeping the "User Mode" and "Enforce Schema Validation" as separate concerns so that's where I started. But then I started writing up the Apex Doc comments on why you'd want to leave Schema Validation ON if you WEREN'T enabling FLS and it sounded silly ("Turn this feature on if you need us to validate that you have syntactical errors in your SOQL"). Why would you want to expend computation expense each time you run a query just because you're sloppy?

So I reasoned that we only need to do this heavy lifting if the legacy form of FLS enforcement was enabled.

This change causes many failing tests .. a few for reasons that are actually silly. But there are a few actual logic changes if you turn schema validation off. Notably, standard objects that have a lookup to a parent usually have an "<parent object>Id" field 

For example, this test fails: 

```
13:35:41.189 (1191507628)|FATAL_ERROR|System.AssertException: Assertion Failed: Expected: SELECT CreatedBy.ManagerId, CreatedBy.Name, FirstName, Id, LastModifiedBy.Email, LastName FROM User, Actual: SELECT CreatedBy.Manager, CreatedBy.Name, FirstName, Id, LastModifiedBy.Email, LastName FROM User

Class.fflib_QueryFactoryTest.deterministic_toSOQL: line 555, column 1
```

We do black magic to transpose the "CreatedBy.Manager" into "CreatedBy.Manager**Id**" 

See https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/master/sfdx-source/apex-common/main/classes/fflib_SObjectDescribe.cls#L110

For that reason, we can't introduce this kind of change as backwards-compatible. 

There are other problems too -- the Set of fields (`private Set<String> fields`) is a case-sensitive Set of strings .. so the Set doesn't implicitly de-dupe someone selecting a field using two different cases (ex "Name" and "name" are different elements in the Set and you end up with a 'duplicate field selected' SOQL error)

So now I'm back at trying to figure out how to accomplish the goal of increasing efficiency of query construction (i.e. avoiding the Describe loops) without creating immense challenges for people adopting this change.



</blockquote></details>

By making the condition more correct (i.e. only bypass the SObjectDescribe stuff if you are opting into the USER_MODE and SYSTEM_MODE), this is now a backwards compatible change. I consider this thread resolved. 
___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 301 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJtTn364S6XMSi_Si9:-NI9FmGF9D5-vUaNDdT3:blg21gk) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/899c3257ac42eb1b4029950568eaa0dfc5082eef/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L301)):*

Yes, will do
</blockquote></details>

Done (see above and below for DRYing)
___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 318 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJv5J-4O4Ax_dyoKC3:-NI9Ftyo8IuH3GtdEEpH:b-896fix) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/899c3257ac42eb1b4029950568eaa0dfc5082eef/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L318)):*
<details><summary><i>Previously, tfuda (Thomas Fuda) wrote…</i></summary><blockquote>

Same comment as above; can't this delegate to the selectField(Schema.SObjectField) method to DRY the code a bit?
</blockquote></details>

Done.
___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 164 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NI9EEdT8mQ8tzTWZB5f:-NI9EEdT8mQ8tzTWZB5g:b-pih4pr) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/13414810a34f6c2d3ffb0044e7a8c73091cf74ff/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L164)):*
> ```vb6
> 			throw new InvalidFieldException('Invalid field: null');		
> 		}
> 		return field.getDescribe().getLocalName();
> ```

ATTENTION: This should be a no-op change. If you're running a SOQL query in a managed package, getting the namespaced name of a field in your namespace is unnecessary

Without this change, you can end up with duplicate fields because you might select the `Amount__c` field on your custom SObject using the SObjectField token (ex. `MyObject__c.Amount__c`) which would return `MyPackage__Amount__c` which would duplicate the same field if you just called `selectField('Amount__c');` (the String overload) 

Like I said above, this change is benign in my namespaced scratch org so I have high confidence in this change
___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 266 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NI9FXWkANamG6bzr88x:-NI9FXWkANamG6bzr88y:b2mlkn5) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/13414810a34f6c2d3ffb0044e7a8c73091cf74ff/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L266)):*
> ```vb6
> 	**/
> 	public fflib_QueryFactory selectField(Schema.SObjectField field){
> 		return selectFields(new Set<Schema.SObjectField>{field});
> ```

the introduction of the `addField()` private method below called attention to how much duplicate code we had. Rather than layer it on thicker, I DRYed this up 
___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 354 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NI9FytUAETwKE6NzTUh:-NI9FytUAETwKE6NzTUi:b5ujmaj) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/13414810a34f6c2d3ffb0044e7a8c73091cf74ff/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L354)):*
> ```vb6
> 		 */
> 		if(mFlsEnforcement == FLSEnforcement.SYSTEM_MODE || mFlsEnforcement == FLSEnforcement.USER_MODE){
> 			fieldPath = fieldPath.toLowerCase();
> ```

As called out above in the now-resolved comment thread in getFieldPath(), we need to normalize the field path to dedupe the Set -- and rather than sprinkle 'toLowerCase()' all over the place, I created this method and then DRYed things up above so we only call it from a couple of places
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls` line 125 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tSvmP4sc2j-kr0CJa:-NI9H2vlEIquR5VeI-yU:b-15pc6h) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/4c180b1c456d8c38dee3b1f5a46266d2386a4ab1/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls#L125)):*

And we need to talk about if/how we support this from `fflib_Application`. 

There is already a newInstance method to construct a SOUOW and pass it an IDML implementation. Adding a "userMode" overload of newInstance() felt wrong. If you want a UserMode version of SOUOW, you should just use the existing newInstance() method and pass it an instance of this class. 
</blockquote></details>

This thread is unresolved but it may not be a blocker. Being able to override the IDML implementation via the constructor overload is existing behavior so for folks that already implemented CRUD/FLS enforcement in SOUOW, they already have a solution that works for generating an SOUOW instance with their IDML instance

I considered adding a `setDML(IDML dmlImpl)` to SOUOW but I don't like introducing the ability to change behavior of an instance after it's constructed and potentially loaded with records

Another option would be to pass an IDML instance to commitWork() via a new override (ex `commitWork(IDML dmlImpl)`)

stohn777 (dismissed):

stohn777:
As expected, unit test "fflib_SObjectSelectorTest.testPolymorphicSelectWithRelatedType" still fails when multi-currency enabled.  [PR 429](https://apexenterprisepattern.slack.com/archives/G01BJ72CLE9/p1670099022814819) should resolve.

tfuda (commented):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)* status: all files reviewed, 10 unresolved discussions (waiting on @daveespo and @ImJohnMDaniel)

___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 314 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJulHk2Xa44kNyYZZu:-NIgQuec1jrTAigjAy9_:b-ozhn7t) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/899c3257ac42eb1b4029950568eaa0dfc5082eef/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L314)):*
<details><summary><i>Previously, tfuda (Thomas Fuda) wrote…</i></summary><blockquote>

Should be "the LIST of {@link Schema.SObjectField}s to select."
</blockquote></details>

You missed this comment on the ApexDoc header block. The param specification is incorrect. The parameter is a List, not a Set.
___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 308 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NIgROvBC1gQ8u2NfMls:-NIgROvBC1gQ8u2NfMlt:b-o20bt4) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/13414810a34f6c2d3ffb0044e7a8c73091cf74ff/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L308)):*
> ```vb6
> 	}
> 
> 	private fflib_QueryFactory selectSObjectField(Iterator<Schema.SObjectField> iter){
> ```

Shouldn't this be named "selectSObjectFields" (plural)?
___
*[`sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls` line 376 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJwGlk0Jfx7rNz20su:-NIgSQdX8g1YeCpJLdYm:bpx0yhk) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/07cd4b54b470108b6fa034b3d2aeb0614d868781/sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls#L376)):*
<details><summary><i>Previously, tfuda (Thomas Fuda) wrote…</i></summary><blockquote>

Minor nit; the method under test is called toSOQL, but your test methods are called toSoql_XXX
</blockquote></details>

Seems like you skimmed over my "non-blocking" comments, so I'm making them blockers.

daveespo (commented):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)* status: 2 of 5 files reviewed, 10 unresolved discussions (waiting on @ImJohnMDaniel, @stohn777, and @tfuda)

___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 314 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJulHk2Xa44kNyYZZu:-NJjpPhHB1wcti54zz9d:b-896fix) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/899c3257ac42eb1b4029950568eaa0dfc5082eef/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L314)):*
<details><summary><i>Previously, tfuda (Thomas Fuda) wrote…</i></summary><blockquote>

You missed this comment on the ApexDoc header block. The param specification is incorrect. The parameter is a List, not a Set.
</blockquote></details>

Done.
___
*[`sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls` line 308 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NIgROvBC1gQ8u2NfMls:-NJjpTDE0V8fagLOwoFF:b-896fix) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/13414810a34f6c2d3ffb0044e7a8c73091cf74ff/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls#L308)):*
<details><summary><i>Previously, tfuda (Thomas Fuda) wrote…</i></summary><blockquote>

Shouldn't this be named "selectSObjectFields" (plural)?
</blockquote></details>

Done.
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls` line 402 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tPeeamhFU-kBqK14r-r3-402:-NJjpXQb65MDc7yIWit8:b-896fix) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/07cd4b54b470108b6fa034b3d2aeb0614d868781/sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls#L402)):*

Admittedly, indent characters are not standard throughout the framework, but since we're touching this function, let's ensure the indent characters are the same for these related functions.
</blockquote></details>

Done.
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls` line 456 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tPeeamhFU-kBqK14r-r3-456:-NJjpZwG-JuDkuRtLdu_:b-896fix) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/07cd4b54b470108b6fa034b3d2aeb0614d868781/sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls#L456)):*

Sync indent character.
</blockquote></details>

Done.
___
*[`sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls` line 376 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NFJwGlk0Jfx7rNz20su:-NJjpaqFCiRE76aUMNB3:b-896fix) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/07cd4b54b470108b6fa034b3d2aeb0614d868781/sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls#L376)):*
<details><summary><i>Previously, tfuda (Thomas Fuda) wrote…</i></summary><blockquote>

Seems like you skimmed over my "non-blocking" comments, so I'm making them blockers.
</blockquote></details>

Done.

stohn777 (approved):

tfuda (commented):
> Reviewed 1 of 3 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)* status: all files reviewed, 7 unresolved discussions (waiting on @ImJohnMDaniel and @stohn777)

sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls
```diff
@@ -121,6 +121,20 @@ public virtual class fflib_SObjectUnitOfWork
 			Database.emptyRecycleBin(objList);
 		}
     }
+
+	public virtual class UserModeDML extends SimpleDML{
```
ImJohnMDaniel, line 125: @daveespo -- Would it not be better to set `UserModeDML` to implement `IDML` instead of extending `SimpleDML`??

daveespo (commented):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)* status: 5 of 6 files reviewed, 8 unresolved discussions (waiting on @ImJohnMDaniel, @stohn777, and @tfuda)

___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls` line 125 at r6](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tPeeamhFU-kBqK14s-r6-125:-NKxJrqP2eQivL4TIeWM:b-3nzph8) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/e3477152af27ff59d1721829bc7848b4a3ae7412/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls#L125)):*
<details><summary><i>Previously, ImJohnMDaniel (John M. Daniel) wrote…</i></summary><blockquote>

@daveespo -- Would it not be better to set `UserModeDML` to implement `IDML` instead of extending `SimpleDML`??
</blockquote></details>

The reason we created SimpleDML was to provide an easy way to extend default functionality without needing to reimplement method bodies (i.e. DRY)

Each time we added a new method to IDML and consumers of this library upgraded, they were forced to copy and paste a default implementation into their custom IDML implementations. By extending SimpleDML and just overriding the methods you want to modify, it keeps the intent of your changes clear. 

Unrelated: Rather than call this "UserModeDML" and have it only support USER_MODE, should we instead change this class to accept a DataAccess argument like the other classes touched in this PR so that you can specify SYSTEM_MODE or USER_MODE (the difference between SYSTEM_MODE and the default DML operation is that sharing rules are not enforce in SYSTEM_MODE)

ImJohnMDaniel:
@daveespo said:

> The reason we created SimpleDML was to provide an easy way to extend default functionality without needing to reimplement method bodies (i.e. DRY)
> Each time we added a new method to IDML and consumers of this library upgraded, they were forced to copy and paste a default implementation into their custom IDML implementations. By extending SimpleDML and just overriding the methods you want to modify, it keeps the intent of your changes clear.

Ok.  We will leave it that question of inheritance alone.

daveespo (commented):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)* status: 4 of 7 files reviewed, 8 unresolved discussions (waiting on @ImJohnMDaniel, @stohn777, and @tfuda)

___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls` line 125 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tSvmP4sc2j-kr0CJa:-NLMGACc30AlcdiOY6mx:bfuxb8p) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/4c180b1c456d8c38dee3b1f5a46266d2386a4ab1/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls#L125)):*

This thread is unresolved but it may not be a blocker. Being able to override the IDML implementation via the constructor overload is existing behavior so for folks that already implemented CRUD/FLS enforcement in SOUOW, they already have a solution that works for generating an SOUOW instance with their IDML instance

I considered adding a `setDML(IDML dmlImpl)` to SOUOW but I don't like introducing the ability to change behavior of an instance after it's constructed and potentially loaded with records

Another option would be to pass an IDML instance to commitWork() via a new override (ex `commitWork(IDML dmlImpl)`) 


</blockquote></details>

And further, the `fflib_Application.UnitOfWorkFactory` has an overload that takes an `IDML` implementation:

`public virtual fflib_ISObjectUnitOfWork newInstance(List<SObjectType> objectTypes, fflib_SObjectUnitOfWork.IDML dml)`

This would be the recommended way to pass in a UserModeDML for consumers that are using `fflib_Application` pattern
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls` line 125 at r6](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-N6tPeeamhFU-kBqK14s-r6-125:-NLMGeGmCGJ_9BMVG5wh:b-ueh4vg) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/e3477152af27ff59d1721829bc7848b4a3ae7412/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls#L125)):*

The reason we created SimpleDML was to provide an easy way to extend default functionality without needing to reimplement method bodies (i.e. DRY)

Each time we added a new method to IDML and consumers of this library upgraded, they were forced to copy and paste a default implementation into their custom IDML implementations. By extending SimpleDML and just overriding the methods you want to modify, it keeps the intent of your changes clear. 

Unrelated: Rather than call this "UserModeDML" and have it only support USER_MODE, should we instead change this class to accept a DataAccess argument like the other classes touched in this PR so that you can specify SYSTEM_MODE or USER_MODE (the difference between SYSTEM_MODE and the default DML operation is that sharing rules are not enforce in SYSTEM_MODE)
</blockquote></details>

@imjohnmdaniel  per our chat, I added an overloaded constructor for `UserModeDML` that lets you specify `SYSTEM_MODE` 

(and rather than define a new enumeration, I was able to use the Apex AccessLevel enumeration because we are only supporting SYSTEM_MODE and USER_MODE with the new UserModeDML class; no need to have a LEGACY enumerated value)

ImJohnMDaniel (approved):

stohn777 (approved):

daveespo (commented):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420)* status: 4 of 7 files reviewed, 8 unresolved discussions

___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls` line 33 at r8](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/420#-NV5Ttqg19kCMVW9UpGu:-NV5cOHX1ruh8nUneKHd:bgljirs) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/d04b6d7b49db40309e3db206e1d2a37f9cc68647/sfdx-source/apex-common/main/classes/fflib_SObjectSelector.cls#L33)):*
<details><summary><i>Previously, clebrsonn (Cleberson) wrote…</i></summary><blockquote>

Why not use the "FLSEnforcement" variable from the class:
"fflib_QueryFactory" here?
</blockquote></details>

There are 4 enumerated values in `fflib_QueryFactory.FLSEnforcement` -- i.e. there is a `NONE` option that isn't present in this enumeration 

That was to support the somewhat incongruous prior behavior of only enforcing CRUD and not FLS via the two different boolean flags on the legacy Selector constructors (enforceCRUD and enforceFLS arguments)

With native User Mode support, we don't have the option to selectively enforce CRUD but not FLS so we wanted to make it clear to the user of the library that their only options going forward are USER_MODE or SYSTEM_MODE -- with LEGACY being a shim to remain backwards compatible.

PR #423, opened 2022-08-02
Change addError method to public and virtual

wimvelzeboer:
It can be useful to add one error string to all the records of a domain directly from a service class, therefore it needs to be public.
And it is also useful to change its behaviour just like the other addError methods, therefore adding virtual

---
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/423)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_SObjects.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjects.cls
@@ -79,7 +79,7 @@ public virtual class fflib_SObjects
 	 *
 	 * @param message The error message to add to each record
 	 */
-	protected void addError(String message)
+	public virtual void addError(String message)
 	{
 		for (SObject record : getRecords())
 		{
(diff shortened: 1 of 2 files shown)


afawcett:
Hi @wimvelzeboer — thanks again for this PR.

We've been discussing it among maintainers. The use case (e.g. applying a single non-field error across all rows from outside a domain subclass, such as from a service) makes sense.

One concern raised internally is consistency: today fflib_SObjects exposes several protected helpers that apply the same kind of operation to every record in the bag (not only addError(String)). If we only promote addError(String) to public, callers can attach object-level errors from outside the hierarchy but still cannot use the same visibility for the closely related APIs, e.g.:

addError(Schema.SObjectField field, String message)
clearField / clearFields
…which feels asymmetric if we accept the idea that orchestration code may legitimately manipulate this record collection.

Proposal: extend this PR (or a follow-up commit on the same branch) so that the same "bulk mutator" family is aligned with addError(String) — at minimum public virtual on the field-level addError overload and on clearField / clearFields, with virtual kept where it already exists so domain subclasses can still override behavior the same way as for addError(String).

If you're willing to update the PR along those lines, we can take another pass for review. 

Thanks for contributing.

afawcett:
I made the above proposed changes for you @wimvelzeboer 👍

PR #441, opened 2023-02-22
Added a decorator method to get picklist value by label

Javwadazeem:
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/441)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
@@ -63,6 +63,7 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 	public Schema.SObjectType table {get; private set;}
 	@TestVisible
 	private Set<String> fields; 
+    private Set<String> setFieldsByLabel;
 	private String conditionExpression;
 	private Integer limitCount;
 	private Integer offsetCount;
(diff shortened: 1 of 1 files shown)


Javwadazeem (commented):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/441)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Javwadazeem)

sfdx-source/apex-common/main/classes/fflib_QueryFactory.cls
```diff
@@ -727,6 +735,14 @@ public class fflib_QueryFactory { //No explicit sharing declaration - inherit fr
 			}	
 		
 			result += String.join(fieldsToQuery,', ');
```
afawcett: Instead of doing the adjustment on the current SOQL being built - it would be safer to do the adjustment as fieldsToQuery is being built or as a second pass after it.

rvandenassum:
@Javwadazeem thanks for submitting this PR! While I fully understand the context and do think this would be a nice addition, I'm not a big fan of the method-naming and structure. 

- `toLabel` isn't only for picklist values, but also for multi-selects, divisions, currency codes, data categories and recordtype names. Would you be open for something like `selectFieldsTranslated()` or `selectFieldUserLanguage()` or `selectFieldsToLabel`
- Additionally, I wouldn't suggest to create a completely different structure, but rather an overload/extension on the existing selectField(s) methods, as this also allows to keep control of field sequence in the SOQL.

Curious for your thoughts!

PR #446, opened 2023-03-30
Readable adjust

clebrsonn:
removed chained ifs and added map to return messages by operation type

- - -
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/446)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_SecurityUtils.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SecurityUtils.cls
@@ -52,7 +52,7 @@ public class fflib_SecurityUtils
     	private CrudException(OperationType operation, Schema.SObjectType objectType){
     		this.m_operation = operation;
     		this.m_objectType = objectType;
-		Map<String, String> messagesByOperation= new Map<String, String>{OperationType.CREATE=> System.Label.fflib_security_error_object_not_insertable,
+		Map<OperationType, String> messagesByOperation= new Map<OperationType, String>{OperationType.CREATE=> System.Label.fflib_security_error_object_not_insertable,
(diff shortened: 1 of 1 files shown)


sfdx-source/apex-common/main/classes/fflib_SecurityUtils.cls
```diff
@@ -52,14 +52,12 @@ public class fflib_SecurityUtils
...
     		this.m_objectType = objectType;
-    		if(operation == OperationType.CREATE)
-    			this.setMessage(System.Label.fflib_security_error_object_not_insertable);
-    		else if(operation == OperationType.READ)
-    			this.setMessage(System.Label.fflib_security_error_object_not_readable);
-    		else if(operation == OperationType.MODIFY)
-    			this.setMessage(System.Label.fflib_security_error_object_not_updateable);
-    		else if(operation == OperationType.DEL)
-    			this.setMessage(System.Label.fflib_security_error_object_not_deletable);
+		Map<OperationType, String> messagesByOperation= new Map<OperationType, String>{OperationType.CREATE=> System.Label.fflib_security_error_object_not_insertable,
```
ImJohnMDaniel, line 55: @clebrsonn this is more of a functional change as opposed to a change for readability.  Was that the intent?  If so, what is the benefit that you see by doing this?

daveespo:
We have a general rule around not merging PRs that are purely "just different, not better". That is, there is no need for this change: It doesn't fix a bug, it doesn't improve performance materially, etc. Making changes for the sake of change introduces risk unnecessarily. 

Additionally, with the introduction of native User Mode support in #419 , the legacy methods for testing CRUD and FLS should be considered deprecated and so any enhancements to that code will likely be rejected.

Issue #221, opened 2019-04-10
UnitOfWork: RegisterDirty with new parent relationship not working


calemcme:
The issue: UnitOfWork does not resolve the relationship at update. Here is the problem description:

Use "registerDirty" with a relationship to a parent record which will be inserted. 
Method to use: 
public void registerDirty(SObject record, Schema.sObjectField relatedToParentField, SObject relatedToParentRecord)

Example: 
create a new transaction "tranNew" record, set the new transaction Id in an existing record "updTran". Code snippet:
mUow.registerNew(tranNew);
mUow.registerDirty(updTran, Transaction__c.NewTransaction__c, tranNew);

After UnitOfWork is committed, the parent field is not filled in record "updTran". The implementation of "updateDmlByType" does not have relationship resolution:

private void updateDmlByType()
{
	for (Schema.SObjectType sObjectType : m_sObjectTypes)
	{		m_dml.dmlUpdate(m_dirtyMapByType.get(sObjectType.getDescribe().getName()).values());
	}
}

It should resolve the relationship before update. The updated implementation will have "m_relationships.get(sObjectType.getDescribe().getName()).resolve();" before update is performed. 

private void updateDmlByType()
{
	for (Schema.SObjectType sObjectType : m_sObjectTypes)
	{
m_relationships.get(sObjectType.getDescribe().getName()).resolve();		m_dml.dmlUpdate(m_dirtyMapByType.get(sObjectType.getDescribe().getName()).values());
	}
}

The modification is similar to the resolution applied in method "private void insertDmlByType()".

private void insertDmlByType()
{
	for (Schema.SObjectType sObjectType : m_sObjectTypes)
	{
m_relationships.get(sObjectType.getDescribe().getName()).resolve();
m_dml.dmlInsert(m_newListByType.get(sObjectType.getDescribe().getName()));
	}
}

calemcme:
The resolution added in "updateDmlByType" may cause an issue if a resolution was resolved at "insertDmlByType". An additional test is necessary to avoid this case.

public void resolve()
 {
            //check relationship Id due to update use.
            if (this.Record.get(this.RelatedToField)==null) {
                this.Record.put(this.RelatedToField, this.RelatedTo.Id);
            }
 }

cnaccio:
[...]

PR #449, opened 2023-04-28
bugfix for Issue 221

jprichter:
fix for https://github.com/apex-enterprise-patterns/fflib-apex-common/issues/221 from @dsurfleet

- - -
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/449)

### DIFF
--- a/sfdx-source/apex-common/test/classes/fflib_SObjectUnitOfWorkTest.cls
+++ b/sfdx-source/apex-common/test/classes/fflib_SObjectUnitOfWorkTest.cls
@@ -610,10 +610,10 @@ private with sharing class fflib_SObjectUnitOfWorkTest
     }
 
     @IsTest
-    private static void testRegisterDirtyWithRelationship() {
+    private static void testRegisterDirtyWithNewParentWhenChildTypeIsBeforeParentType() {
         // GIVEN an existing opportunity
         Opportunity existingOpp = new Opportunity(
-            Id = fflib_IDGenerator.generate(Schema.Opportunity.SObjectType),
+            Id = fflib_SObjectsTest.generateTestRecordId(Schema.Opportunity.SObjectType),
             Name = 'Existing Opportunity',
             StageName = 'Closed',
             CloseDate = System.today()
@@ -637,33 +637,47 @@ private with sharing class fflib_SObjectUnitOfWorkTest
         Test.stopTest();
 
         // THEN
-        System.Assert.isTrue(
-            new fflib_MatcherDefinitions.SObjectsWith(
-                new List<Map<SObjectField, Object>>{
-                    new Map<SObjectField, Object>
-                    {
-                        Account.Id => newAccount.Id,
-                        Account.Name => 'New Account'
-                    }
-                }
-            ).matches(mockDML.recordsForInsert),
-            'The new account record does not match'
-        );
+        System.assertEquals(1, mockDML.recordsForInsert.size(), 'Only the new account should be inserted');
+        SObject insertedAccount = mockDML.recordsForInsert.get(0);
+        System.assertEquals(newAccount.Id, insertedAccount.get(Account.Id), 'The new account Id does not match');
+        System.assertEquals('New Account', insertedAccount.get(Account.Name), 'The new account name does not match');
 
         // AND
-        System.Assert.isTrue(
-            new fflib_MatcherDefinitions.SObjectsWith(
-                new List<Map<SObjectField, Object>>{
-                    new Map<SObjectField, Object>
-                    {
-                        Opportunity.Id => existingOpp.Id,
-                        Opportunity.Name => 'Existing Opportunity',
-                        Opportunity.StageName => 'Closed',
-                        Opportunity.AccountId => newAccount.Id
-                    }
-                }
-            ).matches(mockDML.recordsForUpdate),
-            'The opportunity record should be related to the new Account'
+        System.assertEquals(1, mockDML.recordsForUpdate.size(), 'Only the existing opportunity should be updated');
+        SObject updatedOpportunity = mockDML.recordsForUpdate.get(0);
+        System.assertEquals(existingOpp.Id, updatedOpportunity.get(Opportunity.Id), 'The opportunity Id does not match');
+        System.assertEquals('Existing Opportunity', updatedOpportunity.get(Opportunity.Name), 'The opportunity name does not match');
+        System.assertEquals('Closed', updatedOpportunity.get(Opportunity.StageName), 'The opportunity stage does not match');
+        System.assertEquals(newAccount.Id, updatedOpportunity.get(Opportunity.AccountId), 'The opportunity record should be related to the new Account');
+    }
+
+    @IsTest
+    private static void testRegisterNewRelationshipResolvedDuringInsertDoesNotResolveAgainDuringUpdate() {
+        Product2 product = new Product2(
+            Name = 'Test Product',
+            IsActive = true
+        );
+        PricebookEntry pbe = new PricebookEntry(
+            UnitPrice = 10,
+            IsActive = true,
+            UseStandardPrice = false
+        );
+
+        MockDML mockDML = new MockDML();
+
+        Test.startTest();
+        fflib_SObjectUnitOfWork uow = new fflib_SObjectUnitOfWork(MY_SOBJECTS, mockDML);
+        uow.registerNew(product);
+        uow.registerNew(pbe, PricebookEntry.Product2Id, product);
+        uow.commitWork();
+        Test.stopTest();
+
+        System.assertEquals(2, mockDML.recordsForInsert.size(), 'The product and pricebook entry should be inserted');
+        System.assertEquals(0, mockDML.recordsForUpdate.size(), 'No records should be registered for update');
+        System.assertEquals(
+            product.Id,
+            pbe.Product2Id,
+            'The inserted pricebook entry should keep the product relationship resolved during insert'
         );
     }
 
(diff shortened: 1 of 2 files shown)


stohn777:
Is it reasonable to add a unit test method for this scenario?

john-storey-devops (changes requested):
Please add a test covering this scenario for regression-prevention sake.

jprichter:
> Please add a test covering this scenario for regression-prevention sake.

@stohn777 unit test added [here](https://github.com/apex-enterprise-patterns/fflib-apex-common/pull/449/commits/b9288bdc6fdf80981a6e943f06bcf8761823c4c4)

john-storey-devops (commented):
@jprichter 
The new test DOES NOT fail with the preceding version of fflib_SObjectUnitOfWork.

jprichter:
> @jprichter The new test DOES NOT fail with the preceding version of fflib_SObjectUnitOfWork.

Correct. One way to get it to fail before the change and pass after is to commit to the database. Figured that wouldn't fly.  The MockDML, as it stands and is used throughout the unit tests,  make it

ImJohnMDaniel:
> > @jprichter The new test DOES NOT fail with the preceding version of fflib_SObjectUnitOjkl
> Correct. One way to get it to fail before the change and pass after is to commit to the database instead of using the MockDML. Figured that wouldn't fly.

@jprichter, I am a bit confused then by your comment.  I would expect that the test case, applied to the current version of the codebase (not this PR branch) would fail.  Is that a correct assumption or am I missing something?

jprichter:
> @jprichter, I am a bit confused then by your comment. I would expect that the test case, applied to the current version of the codebase (not this PR branch) would fail. Is that a correct assumption or am I missing something?

@ImJohnMDaniel , the scenario is relating an existing record to a new record. One way to test this is to commit to the database so we have an Id from the newly created record. If you run the test like this, we'll see that the unit test fails on the version that's in master and passes in the version that's in the PR. 

The mocking framework does not create fake Ids after mocking the insert. I'm open to guidance on how to better test this scenario using mocking.

stohn777:
Does this test add value then?  My initial guess is that it does not since the errant code passes the test.  Generally we try to avoid actual DML in tests, but some times, it's needed to properly execute the code being tested.  However some gray exists between that and a test of Salesforce functionality exclusively.

If we can craft a test that properly recreates the problem, using actual DML to add value through testing this feature, without simply testing Salesforce functionality.  Otherwise then adding a test that does NOT add value is pointless.

@jprichter @ImJohnMDaniel

sfdx-source/apex-common/test/classes/fflib_SObjectUnitOfWorkTest.cls
```diff
@@ -848,6 +905,43 @@ private with sharing class fflib_SObjectUnitOfWorkTest
...
+        public List<SObject> recordsForInsert = new List<SObject>();
+        public List<SObject> recordsForUpdate = new List<SObject>();
+        public List<SObject> recordsForDelete = new List<SObject>();
+        public List<SObject> recordsForRecycleBin = new List<SObject>();
+        public List<SObject> recordsForEventPublish = new List<SObject>();
+
+        public void dmlInsert(List<SObject> objList)
+        {
+            for (SObject obj : objList) {
+                obj.Id = fflib_IDGenerator.generate(obj.getSObjectType());
```
ImJohnMDaniel, line 919: @ClayChipps and @jprichter -- Please add a null check on `obj` before trying to generate the mock Id.

sfdx-source/apex-common/test/classes/fflib_SObjectUnitOfWorkTest.cls
```diff
@@ -848,6 +905,43 @@ private with sharing class fflib_SObjectUnitOfWorkTest
         }
     }
 
+    private class MockDMLWithInsertIds implements fflib_SObjectUnitOfWork.IDML
```
ImJohnMDaniel, line 908: @ClayChipps and @jprichter -- Any reason why you did not just simply modify the existing `MockDML`??

ClayChipps:
This bug only presents itself when the Unit of Work has SObjects registered in a certain order.

Here is a sample scenario where the current implementation fails:

Lets assume we have two SObjects that we are about to register to a Unit of Work.
```
Account: {
   Name: My New Account
}
```

```
Opportunity: {
            Id = 0063000000D8cuI,
            Name = 'Existing Opportunity',
            StageName = 'Closed',
            CloseDate = System.today()
}
```

We have registered the Unit of Work such that `Opportunity` is registered before `Account`.

We then attempt to perform the following operation:

```
uow.registerNew(newAccount);
uow.registerDirty(existingOpportunity, Opportunity.AccountId, newAccount);
uow.commitWork();
```

This will fail as when the Unit of Work reaches the [insertDmlByType](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/master/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls#L699), it will first process all Opportunity relationship resolutions (even though no Opportunity has been registered new), and then process the insert and resolution of all Account relationships. 

Then when the Unit of Work reaches the [updateDmlByType](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/master/sfdx-source/apex-common/main/classes/fflib_SObjectUnitOfWork.cls#L706), the relationship resolution is not invoked again, and thus the relationship that was registered has never been resolved after the Account insert.

The initial fix was to add a call to `m_relationships.get(sObjectType.getDescribe().getName()).resolve();` inside of the `updateDmlByType` method so that the relationships would be resolved once again once getting to the update phase, where the Account.Id will be populated.

This causes a separate issue that was exposed by the [InvoicingServiceTest](https://github.com/apex-enterprise-patterns/fflib-apex-common-samplecode/blob/master/sfdx-source/apex-common-samplecode/test/classes/service/InvoicingServiceTest.cls#L28). The `PricebookEntry.Product2Id` field can only be set once, and after it is set it cannot be updated, even in memory. Thus, when the second call to `m_relationships.get(sObjectType.getDescribe().getName()).resolve();` is made within the `updateDmlByType` method, the relationship resolution happens again and attempts to update the PricebookEntry.Product2Id field only to fail with an exception as that field is now read-only.

The solution here was to modify the Relationship resolver to remove relationships from the registered relationships list as they are successfully resolved, so that previously resolved relationships are not resolved a second time.

sfdx-source/apex-common/test/classes/fflib_SObjectUnitOfWorkTest.cls
ClayChipps: done

sfdx-source/apex-common/test/classes/fflib_SObjectUnitOfWorkTest.cls
ClayChipps: Updated to use the existing MockDML.

ImJohnMDaniel:
@ClayChipps and @jprichter  -- After more discussion on this PR, we would like to conduct a test of your changes in a real world series of codebases to assess any functional or performance impacts.  I will reach out to you two directly to discuss this in detail.  Thanks again for the help!

ImJohnMDaniel:
@ClayChipps  and @jprichter -- Is this PR still worth it?  If so, can you let us know and also update the branch.  Thanks for the help.

jprichter:
It's been awhile, so here's a recap of the problem and what was most recently changed. I still think this is a worthwhile fix.

The original problem is that `registerDirty(record, relatedToField, relatedToRecord)` can register an existing record that should be updated to point at a newly inserted parent record. `insertDmlByType()` resolves relationships before insert, but `updateDmlByType()` did not resolve relationships before update. If the Unit of Work processes the dirty child type before the new parent type, the first relationship resolution happens before the parent has an Id. The later update then runs without resolving the relationship again, so the dirty record is updated without the lookup populated.

The fix is to resolve relationships in `updateDmlByType()` immediately before `dmlUpdate()`. Current `master` already has `isResolved()` checks in the relationship resolver, so this rebase keeps that mechanism: unresolved relationships can be resolved later during update, while relationships already resolved during insert are not written again.

I also tightened the tests to address the earlier concern that the previous test did not fail against the old implementation:

- `testRegisterDirtyWithNewParentWhenChildTypeIsBeforeParentType` uses `MockDML` with synthetic Id assignment on insert. It registers `Opportunity` before `Account`, registers a new `Account`, and registers an existing dirty `Opportunity` related to that new account. This fails on `master` without the fix because `Opportunity.AccountId` remains unset, and passes with the fix.
- `testRegisterNewRelationshipResolvedDuringInsertDoesNotResolveAgainDuringUpdate` covers the double-resolution concern raised in the PR discussion, using `PricebookEntry.Product2Id` as the representative path. It verifies that a relationship resolved during insert remains resolved and is not processed again during the update phase.

Updated the branch against current `master` and added regression coverage for the issue. I verified this explicitly in a scratch org:

- `origin/master` + the new tests: `33 ran, 32 passed, 1 failed`

```bash
Failed tests:
fflib_SObjectUnitOfWorkTest.testRegisterDirtyWithNewParentWhenChildTypeIsBeforeParentType: System.AssertException: Assertion Failed: The opportunity record should be related to the new Account
```

- fixed branch + the same tests: `33 ran, 33 passed, 0 failed`


All of this is mock-driven and avoids DML for the regression coverage.

john-storey-devops (approved):

afawcett:
Thanks @jprichter - this is on the agent for our core fflib leads meeting tomorrow.

john-storey-devops (approved):

afawcett:
@jprichter small update - our monthly got bumped to Thursday this week... thanks for your patience.

jprichter:
@afawcett appreciate the updates!

daveespo (approved):
> @daveespo reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/449)* status: all files reviewed, 2 unresolved discussions (waiting on ImJohnMDaniel).

Issue #451, opened 2023-05-06
Selected tests in fflib_SObjectSelectorTest fail in an org with encryption enabled on Account.Name


nicholas-sullivan:
**Describe the bug**
The following tests in fflib_SObjectSelectorTest fail in orgs with Shield Encryption enabled on Account/Opportunity Name.
- toSOQL_When_SystemModeAndChildRelationship_Expect_WellFormedSOQL
- testSOQL_defaultSorting
- toSOQL_When_SystemModeAndParentRelationshipAndDuplicateFields_Expect_WellFormedSOQL

Both of these tests make assertions that the query will contain.

`FROM Account ORDER BY Name`

However getOrderBy() in fflib_SObjectSelector contains the following logic.

```
Schema.SObjectField nameField = describeWrapper.getNameField();
if (nameField != null && !nameField.getDescribe().isEncrypted()) {
  m_orderBy = nameField.getDescribe().getName();
} else {
  m_orderBy = DEFAULT_SORT_FIELD;
```
Where DEFAULT_SORT_FIELD  is set to 'CreatedDate'

This means for orgs without the Name field encrypted the generated SOQL will order by Name otherwise ordering by CreatedDate.

As the tests themselves are hardcoded to use Name in the assertion.

**To Reproduce**

- Create a scratch org with platform encryption enabled
- Enable Shield Encryption on Account Name

Run the following tests from fflib_SObjectSelectorTest.

- toSOQL_When_SystemModeAndChildRelationship_Expect_WellFormedSOQL
- testSOQL_defaultSorting

**Expected behavior**

Expected that both tests pass, instead both fail, instead. Disabling encryption on Account.Name will cause the tests to pass.

System.AssertException: Assertion Failed: 

Expected: 

SELECT name, id, annualrevenue, accountnumber,(.*)\\(SELECT name, id, amount, closedate FROM Opportunities ORDER BY Name ASC NULLS FIRST \\)  FROM Account WITH SYSTEM_MODE ORDER BY **Name** ASC NULLS FIRST

Actual:

SELECT name, id, annualrevenue, accountnumber, (SELECT name, id, amount, closedate FROM Opportunities ORDER BY Name ASC NULLS FIRST )  FROM Account WITH SYSTEM_MODE ORDER BY **CreatedDate** ASC NULLS FIRST

**Version**
Latest Release

daveespo:
@nicholas-sullivan thanks for the fix!

[...]

PR #452, opened 2023-05-06
Bugfix For Issue: 451 - Allow for either Name or CreatedDate as order by.

nicholas-sullivan:
#451 Harcoded Order by name fails with encrypted name.

Add fix so that selected fflib_SObjectSelectorTest  tests pass in orgs where platform encryption is enabled on Account.Name.

Existing tests
- toSOQL_When_SystemModeAndChildRelationship_Expect_WellFormedSOQL
- testSOQL_defaultSorting

Fail in orgs with encryption enabled on Account.Name

The following code in fflib_SObjectSelector means the order field varies from Name in  orgs where name isn't encrypted to CreatedDate where Name is encrypted.

```
 public virtual String getOrderBy()
 {
    if (m_orderBy == null)
    {
        Schema.SObjectField nameField = describeWrapper.getNameField();
        if (nameField != null && !nameField.getDescribe().isEncrypted())
        {
            m_orderBy = nameField.getDescribe().getName();
        }
        else
        {
            m_orderBy = DEFAULT_SORT_FIELD;
            try {
                if (describeWrapper.getField(m_orderBy) == null)
                {
                    m_orderBy = SF_ID_FIELD;
                }
            }
```

Updating the test regex to (?:Name|CreatedDate) ensures both  tests pass in orgs regardless of the encryption status of Account.Name

- - -
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/452)

### DIFF
--- a/sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls
+++ b/sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls
@@ -27,6 +27,10 @@
 @IsTest
 private with sharing class fflib_SObjectSelectorTest 
 {
+	//Get the correct sort field depending on sortable status of Name field
+	static String getDefaultSortField(DescribeSObjectResult sobjDesc) {
+		return sobjDesc.fields.getMap().get('name').getDescribe().isSortable() ? 'Name' : 'CreatedDate';
+	} 
 	@TestSetup
 	static void testSetup(){
 		fflib_SecurityUtilsTest.testSetup();
@@ -167,7 +171,8 @@ private with sharing class fflib_SObjectSelectorTest
 	{
 		Testfflib_SObjectSelectorDefaultSorting selector = new Testfflib_SObjectSelectorDefaultSorting(false);
 		String soql = selector.newQueryFactory().toSOQL();
-		Pattern p = Pattern.compile('SELECT (.*) FROM Account ORDER BY (?:Name|CreatedDate) ASC NULLS FIRST ');
+		Pattern p = Pattern.compile(String.format('SELECT (.*) FROM Account ORDER BY {0} ASC NULLS FIRST ',
+			new List<String>{getDefaultSortField(SObjectType.Account)}));
 		Matcher m = p.matcher(soql);
 		System.assert(m.matches(), 'Generated SOQL does not match expected pattern. Here is the generated SOQL: ' + soql);
 		System.assertEquals(1, m.groupCount(), 'Unexpected number of groups captured.');
@@ -443,7 +448,8 @@ private with sharing class fflib_SObjectSelectorTest
 
 		String soql = sel.createSelectAllWithAccountSOQL();
 
-		String expected = 'SELECT name, id, amount, closedate, account\\.name, account\\.billingpostalcode(.*)FROM Opportunity WITH SYSTEM_MODE ORDER BY Name ASC NULLS FIRST ';
+		String expected = String.format('SELECT name, id, amount, closedate, account\\.name, account\\.billingpostalcode(.*)FROM Opportunity WITH SYSTEM_MODE ORDER BY {0} ASC NULLS FIRST ',
+			new List<String>{getDefaultSortField(SObjectType.Opportunity)});
 		Pattern soqlPattern = Pattern.compile(expected);
 		Matcher soqlMatcher = soqlPattern.matcher(soql);
 		System.assert(soqlMatcher.matches(),'Expected: ' + expected + ' Actual:' + soql);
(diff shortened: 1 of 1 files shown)


stohn777 (changes requested):
Hi @nicholas-sullivan.  Thanks for bringing this to our attention and for the detailed description.

Instead of a blind fallback to the alternate field, let's change the test to expect the CreatedDate field when the Name field is encrypted and not sortable.  Generally I wouldn't want some other cause for Name not being sortable to pass a test that should fail.

I'm open to debate about the condition's details.  Could we confidently rely upon a single field property?  Would being encrypted suffice?

Something like this.  You might want to abstract this to a separate method for DRY purposes.
```
Pattern expectedPattern;
DescribeFieldResult accountNameDfr = Account.SObjectType.getDescribe().fields.getMap().get('name').getDescribe();
if (accountNameDfr.encrypted && !accountNameDfr.sortable)
{
    expectedPattern = Pattern.compile('...'); // The pattern containing "CreatedDate"
}
else
{
    expectedPattern = Pattern.compile('...'); // The pattern containing "Name"
}
```

nicholas-sullivan:
Hi @stohn777 ,

There's also another question regarding the specific usage of isEncrypted() in the getOrderBy() method, it's currently an assumption that encrypted fields can't be sorted however this may change, and hopefully isSortable() should be correct.

Would it better to rely on isSortable() and not isEncrypted() ? Unless there was a specific reason for the encryption check rather than checking it's sortable?

nicholas-sullivan:
Also noticed the same issue applies to the **toSOQL_When_SystemModeAndParentRelationshipAndDuplicateFields_Expect_WellFormedSOQL** test as well.

As sorting is also done on Opportunity Name which will fail if this field is encrypted.

Given that there are multiple combinations between Account and Opportunity Name fields within a single query I believe it would be better to use String.format to replace the appropriate Account/Opportunity name fields rather than conditionaly changing the whole pattern itself.

stohn777:
Hi @nicholas-sullivan.

I see three other tests are failing too because of a flaw in the `fflib_SObjectSelectorTest.Testfflib_SObjectSelector.getOrderBy` method which specifies "Name" in the order-by clause.  Being related, would you like to provide a fix for that too?

The failing tests and errors are:
`testCRUDOff`
```System.QueryException: field 'Name' can not be sorted in a query call
Class.fflib_SObjectSelector.selectSObjectsById: line 336, column 1
Class.fflib_SObjectSelectorTest.testCRUDOff: line 151, column 1
```

`testQueryLocatorById`
```System.QueryException: field 'Name' can not be sorted in a query call
Class.fflib_SObjectSelector.queryLocatorById: line 349, column 1
Class.fflib_SObjectSelectorTest.testQueryLocatorById: line 86, column 1
```

`testSelectSObjectsById`
```System.QueryException: field 'Name' can not be sorted in a query call
Class.fflib_SObjectSelector.selectSObjectsById: line 336, column 1
Class.fflib_SObjectSelectorTest.testSelectSObjectsById: line 61, column 1
```

stohn777:
Hi again @nicholas-sullivan.

You've been busy!!   I'm not sure you meant to do it, but your commit has many additions to the test class beyond the change in `fflib_SObjectSelectorTest.Testfflib_SObjectSelector.getOrderBy`.  Did you mean to commit all of that?  Looks like you have a bunch of more extensive test code, which all pass by the way, but I'm not sure it's needed by the framework.

❔

nicholas-sullivan:
Hi @stohn777 ,

You're right there's probably more in there than is needed, I've instead updated to use AccountNumber instead which given the names were in the same order as the AccountNumber was didn't require any extra changes to the existing tests.

stohn777 (dismissed):

daveespo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/452)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @ImJohnMDaniel and @nicholas-sullivan)

___
*[`sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls` line 31 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/452#-NVA-e96BUU7dzvBgCOJ:-NVA-e96BUU7dzvBgCOK:b8lepdz) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/7736d00485dcb55a0b8accfdc425be80dbe6b475/sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls#L31)):*
> ```vb6
> {
> 	//Get the correct sort field depending on sortable status of Name field
> 	static String getDefaultSortField(DescribeSObjectResult sobjDesc) {
> ```

I don't like this approach to solving these test failures. It's creating a disparity between the logic implemented in the default computation of the Order By clause in `fflib_SObjectSelector` and the assertions in the test methods below. Said differently: The actual logic in `fflib_SObjectSelector` is using `isEncrypted()` to make its logic switch and this method is using `isSortable()` .. and there's logic to deal with SObjects without a CreatedDate field. This `getDefaultSortField()` doesn't embody that logic either.  

Instead, I think that all of the test methods below that are asserting on the resolved SOQL should be calling the actual `getOrderBy()` on the concrete Selector that they're testing to get the appropriate 
___
*[`sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls` line 452 at r4](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/452#-NVA3crL7xXSYOVr0N8b:-NVA3crL7xXSYOVr0N8c:b-3okfm6) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/7736d00485dcb55a0b8accfdc425be80dbe6b475/sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls#L452)):*
> ```vb6
> 
> 		String expected = String.format('SELECT name, id, amount, closedate, account\\.name, account\\.billingpostalcode(.*)FROM Opportunity WITH SYSTEM_MODE ORDER BY {0} ASC NULLS FIRST ',
> 			new List<String>{getDefaultSortField(SObjectType.Opportunity)});
> ```

per my comment above, you'd just call `sel.getOrderBy()` to get the field to sort on

nicholas-sullivan:
Hi @daveespo ,

This has been updated where applicable however for cases such as.

```
public override String getOrderBy()
{
	return 'AccountNumber DESC, AnnualRevenue ASC NULLS LAST';
}
```

Simply relying on the getOrederBy() result will fail as the generated SOQL will add in NULLS FIRST to the first order by, in those cases the exact value has been used.

daveespo (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/452)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ImJohnMDaniel)

stohn777 (approved):

PR #471, opened 2023-11-16
add to DomainFactory and SelectorFactory instantiation by recordTypeId

kamelhanna:
New simple functionality added to DomainFactory and SelectorFactory; be able to make **new Domain and Selector by RecordTypeId**.

Usecase: In your Org if you have different type of ex: Accounts by recordType, now you can make separate domain and selector of each.

- - -
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/471)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_Application.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_Application.cls
@@ -55,7 +55,7 @@ public virtual class fflib_Application
 		}
 
 		/**
-		 * Returns a new fflib_SObjectUnitOfWork configured with the 
+		 * Returns a new fflib_SObjectUnitOfWork configured with the
 		 *   SObjectType list provided in the constructor, returns a Mock implementation
 		 *   if set via the setMock method
 		 **/
@@ -94,7 +94,7 @@ public virtual class fflib_Application
 			if(m_mockUow!=null)
 				return m_mockUow;
 			return new fflib_SObjectUnitOfWork(objectTypes);
-		}		
+		}
 
 		/**
 		 * Returns a new fflib_SObjectUnitOfWork configured with the
@@ -134,7 +134,7 @@ public virtual class fflib_Application
 		public ServiceFactory() { }
 
 		/**
-		 * Constructs a simple Service Factory, 
+		 * Constructs a simple Service Factory,
 		 *   using a Map of Apex Interfaces to Apex Classes implementing the interface
 		 *   Note that this will not check the Apex Classes given actually implement the interfaces
 		 *     as this information is not presently available via the Apex runtime
@@ -164,7 +164,7 @@ public virtual class fflib_Application
 			// Create an instance of the type implementing the given interface
 			Type serviceImpl = m_serviceInterfaceTypeByServiceImplType.get(serviceInterfaceType);
 			if(serviceImpl==null)
-				throw new DeveloperException('No implementation registered for service interface ' + serviceInterfaceType.getName());	
+				throw new DeveloperException('No implementation registered for service interface ' + serviceInterfaceType.getName());
 			return serviceImpl.newInstance();
 		}
 
@@ -183,6 +183,9 @@ public virtual class fflib_Application
 		protected Map<SObjectType, Type> m_sObjectBySelectorType;
 		protected Map<SObjectType, fflib_ISObjectSelector> m_sObjectByMockSelector;
 
+		protected Map<Id, Type> m_recordTypeIdBySelectorType;
+		protected Map<Id, fflib_ISObjectSelector> m_recordTypeIdByMockSelector;
+
 		/**
 		 * Constructs a simple Selector Factory
 		 **/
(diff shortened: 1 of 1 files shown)


wimvelzeboer:
Hi @kamelhanna,
I do not fully understand why this change is needed. I believe you can already do everything with the existing logic, but then implemented differently.

How about the following:
```
IAccounts accounts = Accounts.newInstance(recordIds);   // Generic Accounts
IPartnerAcounts partnerAccounts = Accounts.newPartnerInstance(recordIds);  // Accounts with a certain recordTypeId
```
The domain would then implement the following static methods:
```
public IAccounts newInstance(Set<Id> ids) 
{
  return (IAccounts) Application.Domain.newInstance(ids);
}
public IAccounts newPartnerInstance(Set<Id> ids) 
{
  return ((IAccounts) Application.Domain.newInstance(ids))
          .selectByRecordTypeId(PARTNER_RECORDTYPE_ID);   // Filter method on the domain
}
```
This design pattern would make more sense, it also avoids having to get the record type Id everywhere where you need the domain instance.

This will also avoid the problem where you would create the domain with record Ids that do not belong to the desired recordtype, in this design they do get queried but filtered out.

kamelhanna:
Hello @wimvelzeboer 
Hope all is well,

The main cause that forced us to go with separate domains and selector by recordTypeId is our **Enterprise Org** has different Account types each Account has separate business needs, from the other hand the Sample Code provided with Apex Common don't cover this use case so if there is a sample code cover this challenge provide it to me , and after research and discussion we recommend to use different domains according recordType.

But if the another best practice way to fit our big business requirements without using recordTypeId, we will use it.

Thanks a lot.

ImJohnMDaniel:
G’day @kamelhanna -- First off, Thanks for submitting the PR.

The Selector and Domain Patterns call for only one Domain and only one Selector per SObject.  Multiple domains and selectors per SObject are discouraged in the pattern based.  In theory, with this change, you could setup differing domains and selectors over the same SObject by record type.  That would in turn force the other consumers of the domains and selectors (i.e. Service tier, Client tier, other Domains, etc.) to have more knowledge over the SObject.  This should be avoided as the Domain and Selector have primary responsibility for the SObject’s business logic (i.e. Domain class) and the SObject’s query paths (i.e. Selector class).

Again, while we appreciate your interest in the framework and the offer to help enhance it, we will not be accepting this PR because it is considered somewhat of an anti-pattern to the Domain and Selector patterns.

Please let us know if you have any questions.

wimvelzeboer:
@kamelhanna 
If its indeed two complete different behaviors, then I would suggest to split the source code entirely into two separate packages. Each with their own Application class and domain/selector implementations. 

I have worked in a similar Enterprise Org before. There we make sure that each triggerHandler/domain/selector classe only touched the records of its own record type.

So, in the trigger we basically did something like the following, (to prevent two triggers)
```
fflib_SObjectDomain.triggerHandler(AccountsTriggerHandler_ONE.class);
fflib_SObjectDomain.triggerHandler(AccountsTriggerHandler_TWO.class);
```

The Constructor class in each domain/TriggerHandler then had a filter build in:
```
	public class Constructor implements fflib_SObjectDomain.IConstructable2
	{
		public fflib_SObjectDomain construct(List<SObject> sObjectList)
		{
			return new AccountsTriggerHandler( filter(sObjectList) );
		}

		public fflib_SObjectDomain construct(List<SObject> sObjectList, SObjectType sObjectType)
		{
			return new AccountsTriggerHandler( filter(sObjectList) );
		}
		private List<SObject> filter(List<SObject> records)
		{
		    // filter goes here.
		}
	}
```
I think we needed to do a few isEmpty checks here and there to make it work.

ImJohnMDaniel:
Actually, @wimvelzeboer and @kamelhanna , having multiple Application factories in an org is also an anti-pattern.  

The idea is that the Application Factory needs to be universally accessible to every app in the org so you don't end up with differing implementations.  There are definitely ways for each application to add domain processes specific to the app or to add additional Selector methods or UnitOfWork bindings, but everything in the org should look to a single Application Factory to supply all of its implementations.

Please let me know if you have questions.

cc: @daveespo @stohn777

kamelhanna:
> @kamelhanna If its indeed two complete different behaviors, then I would suggest to split the source code entirely into two separate packages. Each with their own Application class and domain/selector implementations.
> 
> I have worked in a similar Enterprise Org before. There we made sure that each triggerHandler/domain/selector classe only touched the records of its own record type.
> 
> So, in the trigger we basically did something like the following, (to prevent two triggers)
> 
> ```
> fflib_SObjectDomain.triggerHandler(AccountsTriggerHandler_ONE.class);
> fflib_SObjectDomain.triggerHandler(AccountsTriggerHandler_TWO.class);
> ```
> 
> The Constructor class in each domain/TriggerHandler then had a filter build in:
> 
> ```
> 	public class Constructor implements fflib_SObjectDomain.IConstructable2
> 	{
> 		public fflib_SObjectDomain construct(List<SObject> sObjectList)
> 		{
> 			return new AccountsTriggerHandler( filter(sObjectList) );
> 		}
> 
> 		public fflib_SObjectDomain construct(List<SObject> sObjectList, SObjectType sObjectType)
> 		{
> 			return new AccountsTriggerHandler( filter(sObjectList) );
> 		}
> 		private List<SObject> filter(List<SObject> records)
> 		{
> 		    // filter goes here.
> 		}
> 	}
> ```
> 
> I think we needed to do a few isEmpty checks here and there to make it work.

@wimvelzeboer, In my opinion, the idea of filtering records according recordType is more best practice than do that, really I can't thank you enough for your time.

ImJohnMDaniel:
@wimvelzeboer just to be clear, you're saying that they wanted to take an existing Org and split it into two "managed packages" or "unlocked packages"?

wimvelzeboer:
@ImJohnMDaniel Yes, to two separate locked managed packages. In the process of unraveling the dependencies we created these two Application classes.
In the end they chose to have two separate Orgs..

ImJohnMDaniel:
Well okay.  I do find it curious that they went to the managed package route and not unlocked packages for standard org development; even if the intent is to eventually split to multiple orgs.  The introduction of the namespace adds a complexity that is probably unnecessary and forces subsequent development in the respective orgs to have additional constraints that would be avoided if they stay with unlocked packages.

PR #482, opened 2024-02-22
Fixes #440 - changes the few tests that are exercising CRUD/FLS enforcement to prefer the 'Minimum Access - Salesforce' profile since that is a more reliable Profile in most modern orgs

### LINKED ISSUE #440: Expected a QueryException due to read only user not having access to Opportunity
**Describe the bug**
Getting the following error while running test:

> System.AssertException: Assertion Failed: Assertion failed: Expected a QueryException due to read only user not having access to Opportunity
> Class.fflib_SObjectSelectorTest.toSOQL_When_UserModeAndUserCannnotReadObject_Expect_QueryException: line 384, column 1

**To Reproduce**
1. Update to the latest version of Apex Common
2. Verify (or set) Read permissions on Opportunity on "Read Only" security profile
3. Run fflib_SObjectSelectorTest.toSOQL_When_UserModeAndUserCannnotReadObject_Expect_QueryException 

**Expected behavior**
I would expect the test to pass. Here's where the code should be throwing the QueryException:

    sel.selectSObjectsById(new Set<Id>{fflib_IDGenerator.generate(Opportunity.SObjectType)});

It's worth noting that if I remove Read only access to the Opportunity object on the Read Only security profile, the error is thrown and the test passes. However, we don't want to remove read access to this object. Not sure if this is intended to be setup this way for the test to pass or not.

**Screenshots and text of error observed**

> 15:50:49.864 (1933998943)|FATAL_ERROR|System.AssertException: Assertion Failed: Assertion failed: Expected a QueryException due to read only user not having access to Opportunity
> Class.fflib_SObjectSelectorTest.toSOQL_When_UserModeAndUserCannnotReadObject_Expect_QueryException: line 384, column 1
> 

**Version**
Latest, including PR 438 (manually patched as I'm also affected by multi-currency bug).

daveespo:
My first cut at fixing this bug was to try to find an SObject that the Read Only profile DIDN'T grant Read access to. Well, that was a fool's errand because even though there were some Standard Objects that didn't have Object Read access, the FLS on the object granted Read access to every field on those Standard Objects.

I backed away from trying to find a way to thread that needle and instead applied a bit more of a pragmatic approach. Rather than preferring the 'Read Only' profile and if it doesn't exist, create a Minimum Access user and assign them a well-known set of Perms via a Perm Set, let's just start with that (i.e. always prefer the Minimum Access user)

There are potential regressions here for orgs that don't have the Minimum Access profile which does exist out in the wild. Those orgs will fall back to the 'Read Only' profile .. and it's likely that the same tests that are broken as reported in #440 will continue to be broken. But at least this patch fixes the more common case of people not being able to get a clean test run if they create a brand spanking new Developer Edition org (not scratch org)

I tested this on a clean DE org and on a Scratch Org of edition Developer -- tests pass cleanly in both (and prior to this fix, the tests failed as outlined in #440 in the DE org)

- - -
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/482)

### DIFF
--- a/sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls
+++ b/sfdx-source/apex-common/test/classes/fflib_SObjectSelectorTest.cls
@@ -378,7 +378,7 @@ private with sharing class fflib_SObjectSelectorTest
 	static void toSOQL_When_UserModeAndUserCannnotReadObject_Expect_QueryException(){
 		AccessLevelOpportunitySelector sel = new AccessLevelOpportunitySelector();
 
-		User u = getLimitedReadOnlyUser();
+		User u = getMinimumAccessUser();
 		System.runAs(u){
 			try{
 				System.debug(sel.newQueryFactory().toSOQL());
@@ -399,7 +399,7 @@ private with sharing class fflib_SObjectSelectorTest
 	static void toSOQL_When_SystemModeAndUserCannnotReadObject_Expect_Success(){
 		AccessLevelOpportunitySelector sel = new AccessLevelOpportunitySelector(fflib_SObjectSelector.DataAccess.SYSTEM_MODE);
 
-		User u = getLimitedReadOnlyUser();
+		User u = getMinimumAccessUser();
 		System.runAs(u){
 			sel.selectSObjectsById(new Set<Id>{fflib_IDGenerator.generate(Opportunity.SObjectType)});
 		}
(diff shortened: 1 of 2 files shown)


stohn777 (approved):

ImJohnMDaniel (approved):

foxysolutions:
Thanks for the effort and solution @daveespo , looking good!

PR #496, opened 2024-09-10
Type casting bugfix

### LINKED ISSUE #495: Type casting SObjectField to 'String' throws errors for non-string field types
**Describe the bug**
In the `fflib_SObjects` class, all these methods:

- getStringFieldValues
- getRecordsWithBlankFieldValues
- getRecordsWithAllBlankFieldValues
- getRecordsWithNotBlankFieldValues
- getRecordsWithAllNotBlankFieldValues

are type casting SObjectFields to String. This is fine for all the field types that can be casted to String. However, type casting the number types to a String as all the above methods are doing: `(String) record.get(field)` throws errors like `Invalid conversion from runtime type Integer to String`, or `Invalid conversion from runtime type Decimal to String`

**To Reproduce**
Modify the `generateDomain` method from the `fflib_SObjectsTest` class to include in the Accounts the `NumberOfEmployees` field with a random Integer value and then run the entire test class. The tests for all the above methods will now fail.

1. Create a regular scratch org
2. Deploy ApexMocks and apex-common to it
3. Modify the `fflib_SObjectsTest.cls` to include the `NumberOfEmployees` field with an interger value to one or more of the new Accounts and deploy the changes
4. Run the entire test class
5. Notice error: `System.TypeException: Invalid conversion from runtime type Integer to String`

**Expected behavior**
When the above methods are called for any field no matter its type, they correctly return the promised values or records.

**Screenshots and text of error observed**
Example of modified `generateDomain` method:
```apex
private static DomainAccounts generateDomain()
	{
		DomainAccounts domain = new DomainAccounts(
				new List<Account>
				{
						new Account(Name = 'A', ShippingCountry = 'USA'),
						new Account(Name = 'B', ShippingCountry = 'Ireland'),
						new Account(Name = 'C', ShippingCountry = 'UK', NumberOfEmployees = 10),
						new Account(Name = 'D', ShippingCountry = ''),
						new Account(Name = 'E', NumberOfEmployees = 20),
						new Account(),
						new Account(Name = 'G', ShippingCountry = 'Canada', NumberOfEmployees = 30)
				}
		);
		return domain;
	}
```
and the error for one of the existing tests (more are needed to be added to show the extent of the failures)

```bash
fflib_SObjectsTest.itShouldReturnRecordsWithAllFieldValues                    fflib_Objects       Fail     14%      System.TypeException: Invalid conversion from runtime type Integer to String                            
                                                                                                                    Class.fflib_SObjects.getRecordsWithAllNotBlankFieldValues: line 340, column 1                           
                                                                                                                    Class.fflib_SObjectsTest.DomainAccounts.selectPopulatedRecords: line 343, column 1                      
                                                                                                                    Class.fflib_SObjectsTest.itShouldReturnRecordsWithAllFieldValues: line 141, column 1                    8
```

**Version**
Did you try to reproduce the problem against the latest fflib-apex-common code?
Yes,

cnwork:
Fixes #495

Replace the manual type casting of field values into String with `String.valueOf()`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/496)

### DIFF
--- a/sfdx-source/apex-common/main/classes/fflib_SObjects.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjects.cls
@@ -179,7 +179,7 @@ public virtual class fflib_SObjects
 		Set<String> result = new Set<String>();
 		for (SObject record : getRecords())
 		{
-			result.add(String.valueOf(record.get(field)));
+			result.add((String) record.get(field));
 		}
 		return result;
 	}
--- a/sfdx-source/apex-common/main/classes/fflib_SObjects.cls
+++ b/sfdx-source/apex-common/main/classes/fflib_SObjects.cls
@@ -179,7 +179,7 @@ public virtual class fflib_SObjects
 		Set<String> result = new Set<String>();
 		for (SObject record : getRecords())
 		{
-			result.add((String) record.get(field));
+			result.add(String.valueOf(record.get(field)));
 		}
 		return result;
 	}
@@ -254,7 +254,7 @@ public virtual class fflib_SObjects
 		{
 			for (SObjectField field : fields)
 			{
-				if (String.isNotBlank((String) record.get(field))) continue;
+				if (String.isNotBlank(String.valueOf(record.get(field)))) continue;
 
 				result.add(record);
 				break;
@@ -276,7 +276,7 @@ public virtual class fflib_SObjects
 			Boolean allBlank = true;
 			for (SObjectField field : fields)
 			{
-				if (String.isNotBlank((String) record.get(field)))
+				if (String.isNotBlank(String.valueOf(record.get(field))))
 				{
 					allBlank = false;
 					break;
@@ -311,7 +311,7 @@ public virtual class fflib_SObjects
 		{
 			for (SObjectField field : fields)
 			{
-				if (String.isNotBlank((String) record.get(field)))
+				if (String.isNotBlank(String.valueOf(record.get(field))))
 				{
 					result.add(record);
 					break;
@@ -334,7 +334,7 @@ public virtual class fflib_SObjects
 			Boolean allNonBlank = true;
 			for (SObjectField field : fields)
 			{
-				if (String.isBlank((String) record.get(field)))
+				if (String.isBlank(String.valueOf(record.get(field))))
 				{
 					allNonBlank = false;
 					break;
--- a/sfdx-source/apex-common/test/classes/fflib_SObjectsTest.cls
+++ b/sfdx-source/apex-common/test/classes/fflib_SObjectsTest.cls
@@ -121,6 +121,13 @@ public class fflib_SObjectsTest
 
 		System.assertEquals(3, domain.selectWithoutShippingCountry().size());
 	}
+	@IsTest
+	static void itShouldReturnRecordsWithoutFieldValuesForNonStringFields()
+	{
+		DomainAccounts domain = generateDomain();
+
+		System.assertEquals(4, domain.selectWithoutNumberOfEmployees().size());
+	}
 
 	@IsTest
 	static void itShouldReturnRecordsWithoutAllFieldValues()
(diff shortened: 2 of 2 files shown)


ImJohnMDaniel (dismissed):

sfdx-source/apex-common/main/classes/fflib_SObjects.cls
```diff
@@ -257,7 +257,7 @@ public virtual class fflib_SObjects
 		{
 			for (SObjectField field : fields)
 			{
-				if (String.isNotBlank((String) record.get(field))) continue;
+				if (String.isNotBlank(String.valueOf(record.get(field)))) continue;
```
stohn777, line 260: @ImJohnMDaniel  You're a stickler on using explicit blocks for if statements.  I think that would be a good measure here too, especially since I had to look at it several time to see why this method wasn't doing the opposite of its name.  😇

john-storey-devops (commented):
Approving the logic; suggesting change for readability.

daveespo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/496)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @cnwork)

___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 185 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/496#-OD7UHhID-1XLKoKKKld:-OD7UHhID-1XLKoKKKle:bde0j8z) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/d1e8a56c052d28ed54dce52d2b0e5a8a947e7232/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L185)):*
> ```vb6
> 		for (SObject record : getRecords())
> 		{
> 			result.add(String.valueOf(record.get(field)));
> ```

I don't know if I agree that this method should return successfully if the SObjectField is for a non-String field. The coercion from a non-String to a String is an ambiguous behavior (DateTime comes to mind .. since the String version is local time to the running user's time zone). 

The other changes below where it's testing the 'blankness' (which includes null) is less troublesome. 

I would like this change reverted.

daveespo (commented):
Hi @cnwork,

We discussed this thread internally and are going to reject this change for the same reason I stated in my prior comment. While I appreciate the convenience you're looking for with using this method for any field, its behavior is unpredictable and doesn't belong in the core library. It was not the intent of the author of this method to coerce non-Strings into Strings. Its intent was to get the unique set of String values from the records in the Domain.

We expect a ClassCastException if you attempt to call this method and pass a SObjectField that is not a Text type. 

We'd appreciate if you could revert this one change and the rest of the changes are good to be approved.

Thanks!

> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/496#-:-OEL7HOI5XCAI5e9CeP_:bfy1mww)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @cnwork and @ImJohnMDaniel)

daveespo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/496)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @cnwork and @ImJohnMDaniel)

___
*[`sfdx-source/apex-common/test/classes/fflib_SObjectsTest.cls` line 188 at r2](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/496#-OELUxTUCcFSZ4YkikFW:-OELUxTUCcFSZ4YkikFX:b-91683x) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/f493b7236c32770d5dea30704091652b5c772356/sfdx-source/apex-common/test/classes/fflib_SObjectsTest.cls#L188)):*
> ```vb6
> 
> 	@IsTest
> 	static void itShouldReturnStringFieldValuesForNonStringFields()
> ```

@cnwork - sorry, didn't realize this test method was coupled to the change to getStringFieldValues 

Rather than revert the entire test method, let's just rename it and have it confirm that a ClassCastException is thrown .. something like 

```
static void getStringFieldValuesShouldThrowClassCastExceptionForNonStringFields{
 ....

 try{
    domain.getStringFieldValues(Schema.Account.NumberOfEmployees);
    Assert.fail('Expected ClassCastException because NumberOfEmployees is not a Text field');
  }catch(ClassCastException expected){
     //Expected ClassCastException
  }
}
```

afawcett:
Rebased this branch onto current `master` and verified locally in a scratch org:

- `fflib_SObjectsTest` — 18/18 passed
- Full framework test suite — 175/175 passed

The prior review feedback appears addressed: `getStringFieldValues` still expects a cast failure for non-text fields, and the other blankness-check helpers use `String.valueOf()` for non-String field types.

Thanks again @cnwork for the fix and for iterating on the review feedback.

Issue #514, opened 2025-09-03
Code Formatting Analysis and Recommendations


brandbarr:
## Summary

I did some analysis in this repo and wanted to report on the inconsistent code formatting in the `fflib-apex-common` codebase and propose solutions based on community requests in issues [#277](https://github.com/apex-enterprise-patterns/fflib-apex-common/issues/277) and [#491](https://github.com/apex-enterprise-patterns/fflib-apex-common/pull/491).

## Current State Analysis

I analyzed all 32 `.cls` files and found significant formatting inconsistencies across the codebase. Out of 8,933 indented lines total, 6,588 lines (73.7%) use tab indentation while 2,345 lines (26.3%) use space indentation.

The most inconsistent files include `fflib_SObjectUnitOfWork.cls` with 32% tabs and 67% spaces, `fflib_SObjectSelector.cls` with only 5% tabs, and `fflib_SecurityUtils.cls` with just 4% tabs. Meanwhile, most test files use 100% tabs consistently.

The codebase currently follows a modified K&R style with opening braces on the same line as declarations, 4-character indentation (mixing tabs and spaces), and some inconsistent brace placement throughout.

## Proposed Solutions

I see two main approaches we could take:

The minimal change approach would be adding EditorConfig only. This is non-invasive since it doesn't reformat existing code, but prevents future inconsistencies and works across all editors. This directly addresses issue #277.

The complete solution would combine EditorConfig with Prettier. This addresses both issues #277 and #491, provides automatic code formatting, enforces consistent style across all files, and integrates well with CI/CD pipelines. The downside is it requires a one-time mass reformatting and adds the prettier-plugin-apex dependency.

For EditorConfig, the implementation would look like:

```editorconfig
[*.{cls,apex,trigger}]
indent_style = tab
indent_size = 4
tab_width = 4
end_of_line = lf
insert_final_newline = false

[*.json]
indent_style = space 
indent_size = 2
```

For Prettier, the configuration would be:

```json
{
  "printWidth": 120,
  "tabWidth": 4,
  "useTabs": true,
  "semi": true,
  "singleQuote": true,
  "trailingComma": "none",
  "bracketSpacing": true,
  "bracketSameLine": false,
  "endOfLine": "lf",
  "plugins": ["prettier-plugin-apex"],
  "overrides": [
    {
      "files": "*.{cls,apex}",
      "options": {
        "parser": "apex"
      }
    }
  ]
}
```

## Impact Analysis

[...]

PR #521, opened 2025-11-10
Format source with default prettier configuration

nwcm:
Open for discussion, see #514 and #227. I have applied the default prettier configurations for now. I generally lean towards avoiding configuration options where possible as it can become cumbersome to manage with many configured tools. However, that is unlikely to be an issue in this repo

- - -
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/521)


afawcett:
Thanks @nwcm I hope to review another related PR and this sometime in the few months. Thanks for your patience.

PR #536, opened 2026-09-09
fflib_SObjects: 9 new methods, loop optimizations and minor hygiene changes

rvandenassum:
* Added: getRecordIds(field), getRecordIds(field, includeNulls) to return all Ids in a specific lookup field in the records-List
* Added: getRecordIdsFromRelationship(dotPath) to return all Ids from a lookup field on a parent record in the records-List
* Added: getStringFieldValues(field, includeNulls) to return all Strings from a specific field in the records-List
* Added: groupRecordsByFieldValue(field[, toLowerCase]) to group all records in List by a specific SObject field value to a Map<String, List<SObject>>
* Added picklist validation methods: getPicklistValues, getPicklistValuesToStrings, isValidPicklistValue
* Added virtual Exception fflib_Objects.InvalidValueException for centralized, reusable and extendible input validation Exceptions
* Optimised loops from for-each to for-loop leveraging local `numRecords` (note, getRecords() is always initiated as an empty-list, so no null-check needed)
* Added ApexDocs for util methods
* Variable casing rename: SObjectDescribe → sObjectDescribe
* Test: Added 17 test methods to properly cover scenarios of new methods
* Test: Aligned IDGenerator code with apex-mocks for consistency
* Test: Hygiene to have each test method be `private static void` and class itself `private with sharing`

Note: happy to discuss on introduced standardized Exception. Couldn't find one that would match, but open for suggestions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536)


daveespo (changes requested):
I zipped through this quickly tonight -- let Andy do a pass on this and then we can discuss more concretely what changes we'd like to see

> @daveespo reviewed 3 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-:-P1Cwmuu3ouWfcz3LiP1:b-gqqwin)* status: all files reviewed, 4 unresolved discussions (waiting on afawcett and rvandenassum).

___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 50 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1CtCAv1dC2sjcaOGPX:-P1CtCAv1dC2sjcaOGPY:b-ak2ixd) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/90dcd4ce4ead5a48b31ad1f883966d28aaf4055c/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L50)):*
> ```vb6
> 	 * Cached record count of the domain, avoiding repeated calls to getRecords().size() in tight loops
> 	 */
> 	private Integer numRecords
> ```

I don't love this -- I know that we should never add or remove from the original collection passed in -- but we don't have any protections in place to prevent that -- and with this new cached size value, it could lead to some very subtle bugs if someone does use the domain improperly. We even have .setObjects() which would let you update the entire collection 

It seems like we should just call .size() and store the value in a local variable within each method that needs a loop.

(I also wonder if these property getters add their own tax since they seem like they're effectively a method call -- and you're calling the property accessor in the tight loop)
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 149 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1Cs7ls7E52LTAuhSj2:-P1Cs7ls7E52LTAuhSj3:bsn7ovx) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/90dcd4ce4ead5a48b31ad1f883966d28aaf4055c/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L149)):*
> ```vb6
> 		for (Integer i = 0; i < numRecords; i++)
> 		{
> 			Id foundId = (Id) records[i].getSObject(relationshipParts[0])?.get(relationshipParts[1]);
> ```

Can't we support infinite depth here? This is hard-coded to support a single dot (parent) relationship. But it doesn't throw an Exception if the caller provides a grandfather or great-grandfather reference (eg. `Owner.Contact.Account.ParentId` (contrived example, but you get my point)

If it's not reasonable to add infinite depth, we should throw an InvalidValueException if the fieldWithDotRelation contains more than one dot
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 549 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1Cu9Th4z8s2tFOIjnH:-P1Cu9Th4z8s2tFOIjnI:bjc2c7f) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/90dcd4ce4ead5a48b31ad1f883966d28aaf4055c/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L549)):*
> ```vb6
> 	 */
> 	@NamespaceAccessible
> 	public virtual Map<String, List<SObject>> groupRecordsByFieldValue(Schema.SObjectField identifierField, Boolean toLowerCase)
> ```

Why not support Object as the key? What if you want to group by a DateTime or an Id 
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 608 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1Cw-41DXXv4gW6N3wG:-P1Cw-41DXXv4gW6N3wH:b32dq66) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/90dcd4ce4ead5a48b31ad1f883966d28aaf4055c/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L608)):*
> ```vb6
> 	 */
> 	@NamespaceAccessible
> 	public virtual List<Schema.PicklistEntry> getPicklistValues(String objectType, String selectedField)
> ```

Do these picklist methods belong in here? Shouldn't this be part of `fflib_SObjectDescribe` -- it has nothing to do with this specific domain 

Also, these two methods (getPicklistValues and getPicklistValuesToStrings) use Strings for the SObject and Field 

But isValidPicklistValue uses SObjectType and SObjectField 

Why the difference? Can we use SObjectType and SObjectField for all of these new methods?

rvandenassum (commented):
> @rvandenassum made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on afawcett and daveespo).

___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 50 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1CtCAv1dC2sjcaOGPX:-P1HD5DqCpC-URNVj2Xc:b-i80rym) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/151b12e5e60cba4a23f99fd80060aa394c28ed91/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L50)):*

I don't love this -- I know that we should never add or remove from the original collection passed in -- but we don't have any protections in place to prevent that -- and with this new cached size value, it could lead to some very subtle bugs if someone does use the domain improperly. We even have .setObjects() which would let you update the entire collection 

It seems like we should just call .size() and store the value in a local variable within each method that needs a loop.

(I also wonder if these property getters add their own tax since they seem like they're effectively a method call -- and you're calling the property accessor in the tight loop)
</blockquote></details>

@daveespo while I agree with the statement that the records should be untouched, I do agree to move this as a 'real-time' check just before looping the records. Both to avoid the tricky bugs, and possible getter-performance impact.

One thing I'm still considering, is whether we should introduce overloads to allow an input-List, instead of restricting to getRecords(). E.g. if one would first filter by field A, and then want to filter the remaining records by field B. That isn't easily possible now.
However, that would mean an overload for all methods and developers could call new fflib_SObjects().getStringFieldValues( SObjectField, List<SObjects> ) outside of Domain context, which might be in conflict with the SRP of the class.

Curious for your thoughts @daveespo @afawcett .
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 149 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1Cs7ls7E52LTAuhSj2:-P1HIQ2P58jS0Nd2Lcw5:b-896fix) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/151b12e5e60cba4a23f99fd80060aa394c28ed91/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L149)):*

Can't we support infinite depth here? This is hard-coded to support a single dot (parent) relationship. But it doesn't throw an Exception if the caller provides a grandfather or great-grandfather reference (eg. `Owner.Contact.Account.ParentId` (contrived example, but you get my point)

If it's not reasonable to add infinite depth, we should throw an InvalidValueException if the fieldWithDotRelation contains more than one dot
</blockquote></details>

Done.
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 549 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1Cu9Th4z8s2tFOIjnH:-P1HETB55rXnyF59jZ8d:bprrtuw) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/151b12e5e60cba4a23f99fd80060aa394c28ed91/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L549)):*

Why not support Object as the key? What if you want to group by a DateTime or an Id
</blockquote></details>

@daveespo valid question. Reasoning was to keep it simple and avoid complexity of casting upon Map usage. All options are supported, due to the String.valueOf(), however that would indeed mean a developer should first convert to String before calling getter/containsKey on e.g. a Date. 

Personally I'm always inclined to avoid using reference-objects for keys as their likelihood to mismatch while the values are identical, which confuses more junior developers. Converting to String makes that more robust for debugging and consistency in operation. 

Did run a test with Dates and apparently that worked well.
Hence, I think it is more of a preference. I don't want to be too restrictive, but in my perspective converting to Strings makes the usage easier. Though open to adjust.

```
Date d = System.today();
Date d2 = Date.newInstance( 2026, 9, 11);
Map<Date, Boolean> dateMap = new Map<Date, Boolean>{
    d => true
};
System.debug( '*** ' + dateMap.get( d2 ) );
```
output: true
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 608 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1Cw-41DXXv4gW6N3wG:-P1HGZNi6BVvX71_cpv5:biuousp) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/151b12e5e60cba4a23f99fd80060aa394c28ed91/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L608)):*

Do these picklist methods belong in here? Shouldn't this be part of `fflib_SObjectDescribe` -- it has nothing to do with this specific domain 

Also, these two methods (getPicklistValues and getPicklistValuesToStrings) use Strings for the SObject and Field 

But isValidPicklistValue uses SObjectType and SObjectField 

Why the difference? Can we use SObjectType and SObjectField for all of these new methods?
</blockquote></details>

@daveespo this is indeed what I put in our Google Docs as well. 
My conclusion was that fflib_SObjectDescribe is really about doing the Describe calls and handling those values. In these methods we do more than that, and that might cross the single-responsibility; hence felt it would be better fit here. Nevertheless, I fully understand reasoning and am open to move it in `fflib_SObjectDescribe`

Will correct to leverage the stricter types as I agree that makes sense and will make sure to document this for our own backward compatibility

afawcett:
@daveespo @rvandenassum following up on Dave’s comment on `groupRecordsByFieldValue`:

https://github.com/apex-enterprise-patterns/fflib-apex-common/pull/536#pullrequestreview-5173874772

https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1Cu9Th4z8s2tFOIjnH

> Why not support Object as the key? What if you want to group by a DateTime or an Id

I agree with Dave. Prefer `Map<Object, List<SObject>>` so the lookup key is the same type as the field. The current method still uses `String.valueOf`, which pushes a conversion onto every non-text caller:

https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/0268a55bd2e5bd359767f452c57a5e36aba823d8/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L539

Reinier’s `String.valueOf` version (what callers write today):

```apex
Map<String, List<SObject>> byName = domain.groupRecordsByFieldValue(Account.Name);
List<SObject> acme = byName.get('Acme Inc');

Id parentId = someAccount.ParentId;
Map<String, List<SObject>> byParent = domain.groupRecordsByFieldValue(Account.ParentId);
List<SObject> children = byParent.get(String.valueOf(parentId));

Date closeDate = Date.newInstance(2026, 9, 11);
Map<String, List<SObject>> byClose = domain.groupRecordsByFieldValue(Opportunity.CloseDate);
List<SObject> thatDay = byClose.get(String.valueOf(closeDate)); // must match Apex's Date string, typically '2026-09-11'
```

Dave’s `Map<Object, List<SObject>>` (what we should aim for):

```apex
List<SObject> acme = byName.get('Acme Inc');
List<SObject> children = byParent.get(parentId);
List<SObject> thatDay = byClose.get(closeDate);
```

The extra `String.valueOf` on Id/Date is easy to forget, and Id 15 vs 18 or Date format mismatches fail silently with `null`. Apex already treats `Id`, `Date`, `DateTime`, and `Decimal` as value keys — Reinier’s Date snippet in that thread confirmed `dateMap.get(d2)` is `true`. That is the simpler junior API: `byParent.get(parentId)` rather than remember to stringify.

afawcett:
@daveespo @rvandenassum following up on Dave’s comment on the picklist helpers:

https://github.com/apex-enterprise-patterns/fflib-apex-common/pull/536#pullrequestreview-5173874772

https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1Cw-41DXXv4gW6N3wG

> Do these picklist methods belong in here? Shouldn't this be part of `fflib_SObjectDescribe` -- it has nothing to do with this specific domain

I agree with Dave — these should not live on `fflib_SObjects`. `getPicklistValues`, `getPicklistValuesToStrings`, and `isValidPicklistValue` never use the domain’s records:

https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/0268a55bd2e5bd359767f452c57a5e36aba823d8/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L598

Callers can write `new DomainAccounts(accounts).getPicklistValues(Contact.SObjectType, Contact.LeadSource)`, which has nothing to do with those Accounts. That is describe work, and `fflib_SObjectDescribe` is already the cached place for it.

The typed signatures you added after Dave’s note are the right tokens — they should just move with the methods (or we drop this trio from this PR and land the domain helpers without them). Happy either way; I would not add them to the domain public API.

Also worth not carrying `getPicklistValuesToStrings` over as-is: it returns uppercased **labels**, while `isValidPicklistValue` compares **API values**. That mismatch is easy to misuse in translated orgs, and it is another sign this is a describe utility, not a domain operation.

afawcett:
@rvandenassum if your good with the above - i am good to approve - as is @daveespo .

rvandenassum:
@afawcett thanks for the review, suggestions and I agree on the first two suggested changes:
-   Have groupRecordsByFieldValue return Map<Object, List<SObject>> instead of String.valueOf()
-   Move picklistVal to fflib_SObjectDomain

The latter I partially agree, as `getPIcklistValues()` returns a `List<Schema.PicklistEntry>` which is leveraged in both getPicklistValuesToStrings() and isValidPicklistValue(); the first one returns uppercase labels (possibly translated), likely for front-end display; the latter compares API values by ignoring case.

The change made is to rename for clarity:
-   Since the List<String> method doesn't return picklist values, but picklist labels, I've renamed this to `getPicklistLabelsAsStrings`

Lastly, a correction was made to decommission the obsolete SObjectType for PickVal methods, as this is specified already by SObjectField. 

Curious for your response on these changes and to see this moving. Cheers!

afawcett (dismissed):
Approved. Thanks!

daveespo (commented):
> @daveespo made 4 comments and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on afawcett and rvandenassum).

___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 50 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1CtCAv1dC2sjcaOGPX:-P1kN9cF-IagxPgQ4WEo:bxhn884) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/151b12e5e60cba4a23f99fd80060aa394c28ed91/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L50)):*
<details><summary><i>Previously, rvandenassum (Reinier van den Assum) wrote…</i></summary><blockquote>

@daveespo while I agree with the statement that the records should be untouched, I do agree to move this as a 'real-time' check just before looping the records. Both to avoid the tricky bugs, and possible getter-performance impact.

One thing I'm still considering, is whether we should introduce overloads to allow an input-List, instead of restricting to getRecords(). E.g. if one would first filter by field A, and then want to filter the remaining records by field B. That isn't easily possible now.
However, that would mean an overload for all methods and developers could call new fflib_SObjects().getStringFieldValues( SObjectField, List<SObjects> ) outside of Domain context, which might be in conflict with the SRP of the class.

Curious for your thoughts @daveespo @afawcett .
</blockquote></details>

This feels like a bit of a slippery slope with collection manipulation. Today, you'd construct a new instance of the fflib_SObjects for your domain with the already-filtered list .. and then filter again? Seems like that's an OK workaround given the blast radius of internal mutations of the domain collection could have more side effects than we want. 
___
*[a discussion](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1kP1650lEVecYdUTRQ:-P1kP1650lEVecYdUTRR:b94r9ec) (no related file):*
I'm also ready to approve this. I have two open question/comments. Let me know your thoughts and then we can merge this. @rvandenassum 
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjectDescribe.cls` line 290 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1kMqZZ9bdDtxa6KsCJ:-P1kMqZZ9bdDtxa6KsCK:b-keash0) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/468a9ab4e5eee1ccb78bb8723005d41990feafa1/sfdx-source/apex-common/main/classes/fflib_SObjectDescribe.cls#L290)):*
> ```vb6
> 	 */
> 	@NamespaceAccessible
> 	public static List<String> getPicklistLabelsToStrings(Schema.SObjectField selectedField){
> ```

Maybe drop the "ToStrings" part of this method name. Labels are strings .. no conversion is happening.

I don't feel strongly on this one so if you disagree, you can dismiss this request.
___
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 89 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1kOCiH7KzeBgk3YVmq:-P1kOCiH7KzeBgk3YVmr:b-h98kzx) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/468a9ab4e5eee1ccb78bb8723005d41990feafa1/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L89)):*
> ```vb6
> 	 */
> 	@NamespaceAccessible
> 	public virtual Set<Id> getRecordIds(Schema.SObjectField field)
> ```

I didn't see this the first time through .. how does this method differ from the protected `getIdFieldValues` ?

rvandenassum:
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 50 at r1](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1CtCAv1dC2sjcaOGPX:-P1nnDms2tp_I_9t2jqt:boghvq) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/151b12e5e60cba4a23f99fd80060aa394c28ed91/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L50)):*

This feels like a bit of a slippery slope with collection manipulation. Today, you'd construct a new instance of the fflib_SObjects for your domain with the already-filtered list .. and then filter again? Seems like that's an OK workaround given the blast radius of internal mutations of the domain collection could have more side effects than we want.
</blockquote></details>

@daveespo I do agree it requires clear guidelines on how to use them. In my ideation these methods would always construct cloned Lists (as-is) and thus avoid internal domain collection mutations. 

E.g. if one has to process records that are in the Domain records with Stage = 'Closed Won' and Owner being the running user, one can leverage one filter-method, but should still duplicate the logic to filter the remaining records. Honestly, I do think one might consider to create a 'custom local method' which does only one loop to filter directly on both criteria for performance benefits; 

In short: I'm mostly conscious on user-friendliness and re-usability of these methods. Currently, I believe we simply provide a single-filter functionality for the domain, which seems fine. However, it would be a pity if duplications of such methods are scattered across code-repositories for non-Domain setups.

For now not a blocker, and I'll just keep this in mind as I do agree it could be a slippery slope for just having them in fflib_SObjects but not related to the Domain-records-collection.

rvandenassum:
*[`sfdx-source/apex-common/main/classes/fflib_SObjects.cls` line 89 at r3](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1kOCiH7KzeBgk3YVmq:-P1nopzf5j_JQ36sZOCi:b-ef21px) ([raw file](https://github.com/apex-enterprise-patterns/fflib-apex-common/blob/468a9ab4e5eee1ccb78bb8723005d41990feafa1/sfdx-source/apex-common/main/classes/fflib_SObjects.cls#L89)):*

I didn't see this the first time through .. how does this method differ from the protected `getIdFieldValues` ?
</blockquote></details>

@daveespo very sharp! Completely mixed up the input as the getIdFieldValues was introduced later and not yet in our local version before. Then I considered the optimal and clearest method name and came up with this, though we don't want to break signatures of course.

Hence I have:
-   Removed`getRecordIds( Field )`  as duplicate with `getIdFieldValues( Field )`
-   Introduced `includeNulls` overload for getIdFieldValues( Field ) - where the default includes Nulls as-is
-   Renamed `getRecordIdsFromRelationship( String )` to `getParentIdFieldValues( String )`

rvandenassum:

I'm also ready to approve this. I have two open question/comments. Let me know your thoughts and then we can merge this. @rvandenassum
</blockquote></details>

@daveespo @afawcett big thanks for the review, especially given the scarce time with Dreamforce, really appreciated! I've agreed on the method rename and merge; and will keep the idea of opening up Collection Utils in some way to avoid duplications in mind till I have a proper suited solution satisfying both constructs.

rvandenassum (commented):
> @rvandenassum resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on afawcett and daveespo).

daveespo (approved):
> @daveespo reviewed 4 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536)* status: all files reviewed, 1 unresolved discussion (waiting on afawcett).

___
*[a discussion](https://reviewable.io/reviews/apex-enterprise-patterns/fflib-apex-common/536#-P1kP1650lEVecYdUTRQ:-P1oz-3I7oAuGJDQS77l:bm95nol) (no related file):*
<details><summary><i>Previously, rvandenassum (Reinier van den Assum) wrote…</i></summary><blockquote>

@daveespo @afawcett big thanks for the review, especially given the scarce time with Dreamforce, really appreciated! I've agreed on the method rename and merge; and will keep the idea of opening up Collection Utils in some way to avoid duplications in mind till I have a proper suited solution satisfying both constructs.
</blockquote></details>

Thank you for all of the revisions and bringing this to a great spot. Approved and merged!

